Issue #1, opened 2024-01-25
GET /manifest/{projid} route and test


thehabes:
Part of #8 

The application needs to register an endpoint pattern for `/manifest` which handles the GET method.   Code in the support for a pattern like `/manifest/{id}`.  For example:

```
GET
http://localhost:3001/manifest/123
```

Then endpoint will do various bits of logic (TBD at a later time) and ultimately return a JSON Object.

This endpoint does not require authentication and so will not include an Authorization header with a Bearer token.

This route will need CORS headers as described by #9

Tests could include
- Is the route registered with the app
- Does the route process the existence and lack of the {id} section of the request
- When there errors in the request pattern, are they caught and do they elicit a RESTful response
- Does the route return the expected object type
- Does it succeed an end-to-end request

thehabes:
Worth noting that we are not entirely sure whether this will interact with Tiny TPEN, TPEN static, or perhaps need to be able to interact with both.

PR #6, opened 2024-01-31
/manifest route prototype

thehabes:
Relates to #1 and serves as a prototype for how we might set up modular routes and tests.


cubap (approved):
Let's say its good enough to get started

PR #27, opened 2024-02-02
Add CORS.  Add documentation.  Do some cleanup.

thehabes:


manifest/index.mjs
```diff
@@ -13,10 +37,9 @@ export function respondWithManifest(res, manifest){
...
    res.status(200)
    res.json(manifest)
-   res.end()
 }
 
-// Route performs the job
+// Expect an /{id} as part of the route, like /manifest/123
 router.route('/:id')
    .get(async (req, res, next) => {
       let id = req.params.id
```
cubap, line 45: Project ID is most likely going to be a hexadecimal hash, so "is a number" is a bit sneaky.

manifest/manifest.mjs
```diff
@@ -5,11 +5,14 @@ import * as utils from "../utilities/shared.mjs"
 export async function findTheManifestByID(id=null){
    let manifest = null
    if(!utils.validateProjectID(id)) return manifest
+
+   // Mock a pause for endpoints that fail, to mock the time it takes for some async stuff to decide it failed.
    const mockPause = new Promise((resolve, reject) => {
      setTimeout(() => {
        resolve(null)
      }, 3500)
```
cubap, line 13: That's a long pause.

manifest/manifest.mjs
```diff
@@ -5,11 +5,14 @@ import * as utils from "../utilities/shared.mjs"
 export async function findTheManifestByID(id=null){
    let manifest = null
    if(!utils.validateProjectID(id)) return manifest
```
cubap, line 7: Why is `manifest` in here at all? You set to null and then return null.
You probably want to return the validation error (which looks like `false` if I'm reading it right).

manifest/manifest.mjs
```diff
@@ -5,11 +5,14 @@ import * as utils from "../utilities/shared.mjs"
...
+
+   // Mock a pause for endpoints that fail, to mock the time it takes for some async stuff to decide it failed.
    const mockPause = new Promise((resolve, reject) => {
      setTimeout(() => {
        resolve(null)
      }, 3500)
    })
+
    if(id){
       if(id && id===7085) {
```
cubap, line 17: why do we have line 16 and 17?

manifest/manifest.mjs
```diff
@@ -5,11 +5,14 @@ import * as utils from "../utilities/shared.mjs"
...
          manifest = await fetch("https://t-pen.org/TPEN/manifest/7085").then(resp => resp.json()).catch(err => {
```
cubap, line 18: You can just return the fetch. You don't need to ever set `manifest`.

cubap: instead of all these nulls, you can return your mockPause async null simulator.

utilities/shared.mjs
```diff
@@ -1,5 +1,10 @@
...
-// This function is an internal util, but we may want to test it in tests so we export it here for now.
+
+/**
+ * Check if the supplied input is valid JSON or not.
+ * @param input A string or Object that should be JSON conformant.
+ * @return boolean For whether or not the supplied input was JSON conformant.
+ */ 
 export function isValidJSON(input=""){
    if(input){
       try{
```
cubap, line 10: Not sure what's happening here. You are preempting some checks with the default "" set. with a try here, most things are all caught...

```js
function isValidJSON(input) {
  try {
    return Boolean((typeof input === "string") ? JSON.parse(input) : JSON.parse(JSON.stringify(input)))
  } catch(err) {
    return false
  }
}
```

Maybe. I don't know

cubap (approved):
This works, but I did leave a lot of notes.

manifest/index.mjs
thehabes: Fair.  Built it off of current functionality where the ID we get is an Integer Project ID.  Until a decision is made on exactly what it is any 'validation function' is just theoretical.  This is a stub to say "get ready to process a TPEN Project Entity Id here".  I added in the assumption an Integer number would suffice as the minimum, perfectly acceptable if not.  When we know what the ID is going to be, we can come in and update this validation.

manifest/manifest.mjs
thehabes: oo yeah forgot I left it at 3.5 seconds to make sure I noticed it...

manifest/manifest.mjs
thehabes: No.  The function receives a project ID as input and returns null or the matching manifest.  It mocks some Manifest logic.  Given a project ID, go get the Manifest.
```
const manifestObj = await logic.findTheManifestByID(id)
```
Is either null or a Manifest

manifest/manifest.mjs
thehabes: C'mon Han what are you doing.

manifest/manifest.mjs
thehabes: This function does not return a promise.  It returns a JSON object or null.

utilities/shared.mjs
thehabes: String or object comes in.  If it can be parsed to JSON, return true.  Otherwise, return false.  Make sure "" and null values return false.  That's what it does.

PR #32, opened 2024-02-04
WIP on issue 19

Sandeep-1-Kumar:
This issue is to resolve an API request to resource : page
Required URL parameters : Id (as Number)

Added Code : Added Package "page" containing "controller.mjs" and "model.mjs"
Changed Code : updated "app.mjs" adding route to resource "page"

steps a reviewer can take to experience the impact of changes:
-npm install (ignore if already did)
-npm start
##Listening on 3001
URL test (postman/browser) : GET : http://localhost:3001/page/123
No id case : GET : http://localhost:3001/page/
Number Check : http://localhost:3001/page/potato


thehabes (changes requested):
Please only take PRs out of Draft when you feel you have completed all the requirements of your issue.  When you take a PR out of Draft that signals to us you have completed the requirements of your issue and are having us review and test the code for submission and deployment.

There are no tests written.  Please convert this back into Draft write tests for your endpoint.

I was able to run the code and test the endpoint via a browser.
http://localhost:3001/page/11 does not have a good message
http://localhost:3001/page/potato does not have a good message

page/model.mjs
```diff
@@ -0,0 +1,29 @@
+import * as utils from "../utilities/shared.mjs"
+
+export async function findPageById(id = null) {
+    let page = null;
+  
+    if (!utils.validateProjectID(id)) {
```
cubap, line 6: This validation is for a projectID, not a pageID. In this case, it isn't broken yet, but it is important to see that the object validation here is not honored.

page/model.mjs
```diff
@@ -0,0 +1,29 @@
...
+  
+    const pageArray = [
+      { id: 123, text: "welcome to the page" }
+    ];
+  
+    page = pageArray.find((page) => page.id === id);
+  
+    if (page === null) {
+        page = await mockPause.then((val) => {
+        return null;
```
cubap, line 24: `mockPause` returns `null` so you don't have to explicitly return it here.

cubap (changes requested):
Really close. I have not pulled this to test it yet, but will append this when I have.

page/controller.mjs
```diff
@@ -0,0 +1,70 @@
...
+      'X-HTTP-Method-Override'
+    ],
+    exposedHeaders: '*',
+    origin: '*',
+    maxAge: '600'
+  })
+);
+
+
+  router.route('/:id?')
```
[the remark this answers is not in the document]
Sandeep-1-Kumar: I had tested and its working properly , if id is not matched it will fall into 404 and thats what we need!!!

page/model.mjs
Sandeep-1-Kumar: Updated this , please check.

page/model.mjs
Sandeep-1-Kumar: sure, updated.

thehabes (approved):
As we move forward, note that you should not use `;` except when necessary.  Modern javascript does not require that you end your lines with `;`

PR #34, opened 2024-02-05
WIP on Issue 2

leandrumartin:
* Added 'project' directory mirroring the structure of the existing 'manifest' directory, including files and tests
* Added project.json to 'public' directory, with the same dummy data as manifest.json
* Dummy data is loaded from the project.json file instead of the TPEN website

* After running `npm start`, navigating to http://localhost:3001/project/{project_id_here} will test out the functionality.
* As of now, functionality is largely mirrored from the /manifest endpoint. http://localhost/project/7085 loads dummy data from the project.json file, and any other URLs result in appropriate error messages.
* `npm run unitTests` runs the new tests, which currently test the same basic functionality as the tests for /manifest

### DIFF
--- a/project/__tests__/exists_unit.test.mjs
+++ b/project/__tests__/exists_unit.test.mjs
@@ -1,0 +1,16 @@
+//need to import app for coverage, not for actual testing tho.
+import app from '../../app.mjs'
+
+describe('Project endpoint availability unit test (via a check on the app routes). #exists_unit', () => {
+  it('responds to /project/id', () => {
+    let exists = false
+    const stack = app._router.stack
+    for(const middleware of stack){
+      if(middleware.regexp && middleware.regexp.toString().includes("/project")) {
+         exists = true
+         break
+       } 
+    }
+    expect(exists).toBe(true)
+  })
+})
--- a/project/project.mjs
+++ b/project/project.mjs
@@ -1,6 +1,7 @@
 /** Logic for the /project endpoint */
 
 import * as utils from '../utilities/shared.mjs'
+import * as fs from 'fs'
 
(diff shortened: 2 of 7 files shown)


project/__tests__/exists_unit.test.mjs
```diff
@@ -0,0 +1,16 @@
+//need to import app for coverage, not for actual testing tho.
+import app from '../../app.mjs'
+
+describe('Project endpoint availability unit test (via a check on the app routes). #exists_unit', () => {
+  it('responds to /project/id', () => {
+    let exists = false
+    const stack = app._router.stack
+    for(const middleware of stack){
+      if(middleware.regexp && middleware.regexp.toString().includes("/project")) {
```
cubap, line 9: just to note that this match would also match `/projectCopy` or some other accidental collision.

project/project.mjs
```diff
@@ -0,0 +1,34 @@
...
+        console.error(err)
+        return null
+      }
+    })
+    project = projectFileBuffer !== null && JSON.parse(projectFileBuffer.toString())
+  }
+
+  // Mock the scenario where it takes a couple seconds to look for but not find the Project.
+  if (project === null) {
+    project = mockPause.then((val) => {
```
cubap, line 29: `mockPause()` resolves `null` so you can just return it directly.

public/project.json
```diff
@@ -0,0 +1,42 @@
+{
```
thehabes, line 1: Let's delete this file.  I am going to delete my `manifest.json` on my branch as well.

thehabes (approved):
We can get rid of project.json upstream.

PR #45, opened 2024-02-09
Initial Test & Deployment Workflows

thehabes:


cubap (approved):
This all reads right. I can't test it more directly than you already have, so I'll lean on that trust.

thehabes:
Sounds good!  We'll probably get more picky about which tests run, but this should suffice as the starter.

Issue #50, opened 2024-02-10
Initiate Layer Entity Class


cubap:
A module for a "Layer" is needed for use by the other services. In TPEN, a Layer is an ordered list of Annotation Pages that make up a layer of transcription for a Project. The module should support the methods needed for the api services.

A snip from the relevant document includes:

> Single Annotation Page Collection as layer of project
> * Project URL 
> * Sibling layers by reference 
> * Parameterize
>     * Only text blob, page arrays, lines arrays 
>     * Only set of image URLs 
>     * Embed referenced documents
>     * HTML page turner document of a single layer 

The module should be created and importable, but all the methods can just stub out. Tests come first and can simply `expect(methodName).toBeInstanceOf("Function")`.

The clip above is a GET but there is also update and delete things to include.

PR #54, opened 2024-02-19
WIP on Issue #50

leandrumartin:
Added asJSON, create, remove, save, and fetch methods in Layer class. The asJSON method includes a skeleton of an object structure.

Now, all existing tests for the Layer class now pass when running `npm run unitTests`.


classes/Layer/Layer.mjs
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    set id(id) {
+        this.layer.id = id
+    }
+
+    asJSON() {
+        return {
+            type: 'Range',
+            '@context': 'http://www.w3.org/ns/anno.jsonld',
```
cubap, line 20: https://iiif.io/api/presentation/3.0/ Look at the Range here for a "real" example.

cubap: Still incorrect

classes/Layer/Layer.mjs
```diff
@@ -0,0 +1,78 @@
...
+
+    fetch() {
+        return Promise.resolve(new Layer())
+    }
+
+    getSiblingLayers() {
+        return Promise.resolve('Array of all sibling layers in this Layer\'s Manifest')
+    }
+
+    getSiblingLayer(id) {
```
cubap, line 46: This would already be in the Layer constructor

classes/Layer/Layer.mjs
```diff
@@ -0,0 +1,78 @@
...
+
+    getTextBlob() {
+        return Promise.resolve('Text contents of pages')
+    }
+
+    fetchHTMLDocuments() {
+        return Promise.resolve('HTML documents of pages')
+    }
+
+    embedReferencedDocuments() {
```
cubap, line 70: this is about how asJSON() is serialized is its either a parameter up there or a private method.

cubap (changes requested):
minor changes

cubap (approved):

PR #64, opened 2024-03-04
WIP on Issue 57 extend line web api

BhanuKandula5030:
What was changed?

**Updated Endpoint Logic:** Refactored the GET /line/{id} endpoint in the index.js file and also updated the line.mjs file.

**Database Abstraction:** Introduced a database controller abstraction with db.getByID() and db.find({property: match}) methods, simulating the connection to an actual database. 

**Alternate Returns for Query Parameters:** Implemented logic to handle query parameters such as ?text=blob, ?image=full, ?image=line, and ?lookup=project mentioned in the issue. 

**Functionality unit tests :** Implemented the tests to  ensure the different query parameters are working fine.

**Database Mock:** Created a simple database mock (db.js) to simulate database interactions for testing purposes.


line/__tests__/end_to_end_unit.test.mjs
```diff
@@ -1,58 +1,50 @@
-import lineRouter from '../index.mjs'
-import express from 'express'
-import request from 'supertest'
+import lineRouter from '../index.mjs';
```
thehabes, line 1: Watch out for the extraneous `;` chars

line/line.mjs
```diff
@@ -1,27 +1,85 @@
...
+  line = linesArray.find((line) => line.id === id) || null;
 
   if (line === null) {
-    line = await mockPause
+    line = await mockPause;
+  }
+
+  if (line !== null) {
+    if (options.text === 'blob') {
+      return { text: line.textualBody };
```
thehabes, line 32: You will have to set the Content-Type for this.  This should return something that is "text" and not a JSON.  You would set the type to "text/plain" then return the text blob that is the text of ` line.textualBody`

line/line.mjs
```diff
@@ -1,27 +1,85 @@
...
+        textualBody: line.text,
+        project: line.project,
+        canvas: line.canvas,
+        layer: line.layer,
+        viewer: line.viewer,
+        license: line.license
+      };
+
+      if (options.view === 'xml') {
+        return generateXML(line);
```
thehabes, line 54: The return for this does not show an XML page, it shows JSON
image

line/line.mjs
```diff
@@ -1,27 +1,85 @@
...
+        id: line.id,
+        '@type': line['@type'],
+        creator: line.creator,
+        textualBody: line.text,
+        project: line.project,
+        canvas: line.canvas,
+        layer: line.layer,
+        viewer: line.viewer,
+        license: line.license
+      };
```
thehabes, line 51: I do not see this formed JSON come out, I still see the old primitive value.  I should see these values on the returned Line object.
image

thehabes (changes requested):
Make sure the endpoints actually process and return the correct things for some of these parameters you are building in support for.

line/__tests__/end_to_end_unit.test.mjs
BhanuKandula5030: Updated this without ; chars

line/line.mjs
BhanuKandula5030: Updated the code. Thanks for the  comment

line/line.mjs
BhanuKandula5030: Uodated code shows json

leandrumartin (approved):

line/index.mjs
```diff
@@ -1,71 +1,70 @@
-import express from 'express'
-import * as utils from '../utilities/shared.mjs'
-import cors from 'cors'
-import { findLineById } from './line.mjs'
+import express from 'express';
```
cubap, line 1: It looks like you have an auto formatter adding semicolons at the end of all the lines. That makes the diff harder to read and is not the style of the rest of the code. We only use ";" to when required to by the code (and then usually at the start of the offending line instead, for what it is worth).

line/line.mjs
```diff
@@ -1,27 +1,86 @@
...
 
-  line = linesArray.find((line) => line.id === id) || null
+  line = linesArray.find((line) => line.id === id) || null;
 
   if (line === null) {
-    line = await mockPause
+    line = await mockPause;
+  }
+
+  if (line !== null) {
```
cubap, line 30: you can unwrap this `if` your guard clause saved you a whole indent!

line/line.mjs
```diff
@@ -1,27 +1,86 @@
...
+    if (options.text === 'blob') {
+      return { 'Content-Type': 'text/plain', body: line.text };
+    } else if (options.image === 'full') {
```
cubap, line 33: These `else if` make me nuts. This is a great place for a few switch cases to move through the suboptions much faster. You could also check if there are parameters right up front to short circuit this.

line/line.mjs
cubap: I'm not seeing the update. It still looks like you are returning an object instead of setting headers and adding bodies. I'll keep looking and peek through the commits shortly.

cubap (changes requested):
I don't hate it but it needs a little more attention.

line/__tests__/end_to_end_unit.test.mjs
```diff
@@ -1,39 +1,39 @@
-import lineRouter from '../index.mjs';
-import express from 'express';
-import request from 'supertest';
+import lineRouter from '../index.mjs'
```
cubap, line 1: The semicolons are gone and back? You need to make sure you have your workspace settings consistent.

line/line.mjs
```diff
@@ -29,13 +29,14 @@ export async function findLineById(id = null, options = {}) {
...
     } else if (options.image === 'full') {
-      return { image: line.canvas };
+      // Assuming line.canvas contains the full image
+      return { 'Content-Type': 'image/jpeg', body: line.canvas };
     } else if (options.image === 'line') {
-      return { image: `some line image URL for id ${id}` };
+      return { 'Content-Type': 'image/jpeg', body: `some line image URL for id ${id}` };
     } else if (options.lookup) {
-      return { relatedDocument: `some ${options.lookup} document for id ${id}` };
+      return { 'Content-Type': 'text/plain', body: `some ${options.lookup} document for id ${id}` };
```
cubap, line 39: This is still an Object. You want to look at adding the ContentType to the `res` Response here and then the body will be understood by the client when you `.send()` it. You might also want to check the statusCode.

line/index.mjs
BhanuKandula5030: I have updated in the code

line/line.mjs
BhanuKandula5030: updated in the code

line/line.mjs
BhanuKandula5030: I have used the switch cases in the updated code

line/__tests__/end_to_end_unit.test.mjs
BhanuKandula5030: Thanks updated in the code

line/line.mjs
BhanuKandula5030: I have addressed the status codes in the updated code

line/line.mjs
BhanuKandula5030: Addressed using the headers and body in the updated code

line/line.mjs
```diff
@@ -1,27 +1,100 @@
...
+    },
+  ];
+
+  let line = linesArray.find((line) => line.id === id) || (await mockPause);
+
+  if (line === null || line.id !== id) {
+    return { statusCode: 404, body: `TPEN 3 line "${id}" does not exist.` };
+  }
+
+  switch (options.text) {
```
cubap, line 35: A single conditional doesn't need a switch, but this part is all fluid anyway.

line/line.mjs
```diff
@@ -1,27 +1,100 @@
...
+    case 'blob':
+      return { statusCode: 200, headers: { 'Content-Type': 'text/plain' }, body: line.text };
+    default:
+      break;
+  }
+
+  switch (options.image) {
+    case 'full':
+      return { statusCode: 200, headers: { 'Content-Type': 'image/jpeg' }, body: line.canvas };
```
cubap, line 44: This isn't correct, but I think it can be fixed in a separate pass.

cubap (changes requested):
This definitely isn't the last time we'll be through this code, so there are some things that need to be accepted as just incomplete. I would never hold back an entire PR just for semicolons, but it has gone back and forth so many times here that I am worried about your development environment/workflow. 

If you don't address anything else but can confidently remove the ";" that are unneeded, I will endorse pulling this in.

line/line.mjs
BhanuKandula5030: Updated with the if in the code for the blob

line/index.mjs
```diff
@@ -42,10 +40,10 @@ router.route('/:id')
 
       const lineObject = await findLineById(id)
 
-      if (lineObject !== null) {
-        respondWithLine(res, lineObject)
+      if (lineObject.statusCode === 404) {
+        return utils.respondWithError(res, 404, lineObject.body)
       } else {
```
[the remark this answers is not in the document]
BhanuKandula5030: Thanks for the comment .Removed the else part in the updated code

line/__tests__/end_to_end_unit.test.mjs
```diff
@@ -1,58 +1,43 @@
-import lineRouter from '../index.mjs'
-import express from 'express'
-import request from 'supertest'
+import lineRouter from '../index.mjs';
+import express from 'express';
```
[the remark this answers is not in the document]
BhanuKandula5030: I was using the git pull and modified code that might be the reason .i will try to avoid using the formatter and I have made changes in the updated code.

cubap (approved):
I think we're good now.

PR #65, opened 2024-03-04
WIP on Issue #58

leandrumartin:
I added tests for the /project endpoint, testing for appropriate responses to badly formatted project IDs and for a correct code response to a good request. I also implemented better error handling for a better error message in case of a server-side error. The tests can be run with `npm run unitTests`.

### DIFF
--- a/project/index.mjs
+++ b/project/index.mjs
@@ -29,13 +29,100 @@ router.use(
   })
 )
 
-// Send a successful response with the appropriate JSON
+// Send a successful response with the appropriate JSON or alternate response based on params
 export function respondWithProject(res, project) {
   const id = project['@id'] ?? project.id ?? null
-  res.set('Content-Type', 'application/json; charset=utf-8')
-  res.location(id)
-  res.status(200)
-  res.json(project)
+
+  let textType = req.query.text
+  let image = req.query.image
+  let lookup = req.query.lookup
+  let view = req.query.view
+  let responseType = [textType, image, lookup, view]
+    .find(elem => elem !== undefined) // Only one response type is valid
+
+  let embed = req.query.embed
+
+  if (!responseType || responseType === view && view == "json") {
+    res.set('Content-Type', 'application/json; charset=utf-8')
+    res.location(id)
+    res.status(200)
+    res.json(project)
+    return
+  }
+
+  let retVal;
+  switch (responseType) {
+
+    case textType:
+      switch (textType) {
+        case "blob":
+          res.set('Content-Type', 'text/plain; charset=utf-8')
+          // return: a complete blob of text of all lines concatenated
+          retVal = new Blob(
+            project.layers.map(layer =>
+              db.getByID(layer).getLines().map(line => line.textualBody).join(" ")
+            ).join(" ")
+          )
+          break
+        case "layers":
+          res.set('Content-Type', 'application/json; charset=utf-8')
+          retVal = project.layers.map(layer => db.getByID(layer))          
+          break
+        case "pages":
+          res.set('Content-Type', 'application/json; charset=utf-8')
+          retVal = project.layers.flatMap(layer => db.getByID(layer).getPages())
+          break
+        case "lines":
+          res.set('Content-Type', 'application/json; charset=utf-8')
+          break
+        default:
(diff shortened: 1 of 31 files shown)


project/index.mjs
```diff
@@ -0,0 +1,173 @@
...
+// Send a successful response with the appropriate JSON or alternate response based on params
+export function respondWithProject(req, res, project) {
+  const id = project['@id'] ?? project.id ?? null
+
+  let textType = req.query.text
+  let image = req.query.image
+  let lookup = req.query.lookup
+  let view = req.query.view
+  let responseType = [textType, image, lookup, view]
+    .find(elem => elem !== undefined) // Only one response type is valid
```
thehabes, line 41: You will actually want to throw an error here if more than one of this is !==undefined.

project/index.mjs
```diff
@@ -0,0 +1,173 @@
...
+
+  let embed = req.query.embed
+
+  if (!responseType || responseType === view && view == "json") {
```
thehabes, line 45: You will have specifically picked the correct 'responseType' above.  As such, you can remove this code block and go straight into the switch.  This is the `default` case when nothing is provided.  You will handle the JSON view case as a case.

project/index.mjs
```diff
@@ -0,0 +1,173 @@
...
+            project.layers.map(layer =>
+              db.getByID(layer).getLines().map(line => line.textualBody).join(" ")
+            ).join(" ")
+          ) */
+          retVal = new Blob('mock text')
+          break
+        case "layers":
+          res.set('Content-Type', 'application/json; charset=utf-8')
+          /* retVal = project.layers.map(layer => db.getByID(layer)) */
+          retVal = [['mock text'], ['mock text']]
```
thehabes, line 71: This should be an Array of Layer objects like
```json
[
     { "name": "Layer.name", "id": "#AnnotationCollectionId", "textContent": "concatenated blob" }
]
```

project/index.mjs
```diff
@@ -0,0 +1,173 @@
...
+          break
+        case "pages":
+          res.set('Content-Type', 'application/json; charset=utf-8')
+          /* retVal = project.layers.flatMap(layer => db.getByID(layer).getPages()) */
+          retVal = [['mock text'], ['mock text']]
```
thehabes, line 76: This should be an Array of Page objects like

```json
[
     { "name": "Page.name", "id": "#AnnotationPageId", "textContent": "concatenated blob" }
]
```

project/index.mjs
```diff
@@ -0,0 +1,173 @@
...
+          break
+        case "lines":
+          res.set('Content-Type', 'application/json; charset=utf-8')
+          retVal = [['mock text'], ['mock text']]
```
thehabes, line 80: This should be an Array of Page objects like

```json
[
      { "name": "Page.name", "id": "#AnnotationPageId", "textContent": [{ "id" : "#AnnotationId", "textualBody" : "single annotation content" }]}
]
```

project/index.mjs
```diff
@@ -0,0 +1,173 @@
...
+            'Improper request.  Parameter "image" must be "thumbnail."')
+          break
+      }
+      break
+
+    case lookup:
+      switch (lookup) {
+        case "manifest":
+          // return: (layers[], annotations[], metadata[], group) find the related document or Array of documents and return that instead, the version allowed without authentication
+          retVal = [['mock text'], ['mock text']]
```
thehabes, line 108: This will be a resolved JSON manifest.  The value for `manifest` in the return will not be the URI, it will be the actual JSON of the resolved URI.

thehabes (changes requested):
I expected http://localhost:3010/project/7085?view=html to return to me an HTML page containing a project, but got nothing.  Have this return a mock HTML view of a project, for example

``` HTML
<html>
     <body> 
          <pre tpenid="7085"> {"id": "7085", ...}</pre>  
     </body>
</html>
```
Make sure the flags result in something, I am getting a lot of blank pages when testing them.  Make sure that something is of the proper shape and type, the `[['mock text'], ['mock text']]` does not meet the requirements for the mock things you should be returning.

cubap (approved):
I've recorded some quibbles, but I'm okay with this.

PR #66, opened 2024-03-06
59 db driver + controllers + endpoint usage demo

cubap:
Establish an abstract database handler such that regularized calls like `abstract.create(data)` can create in any of the databases that house TPEN3 data.  We currently imagine this as...

1. The RERUM database.  This requires a TinyPen controller/client that uses the TinyPEN web API endpoints.
2. A private MongoDB database for private app data.  This requires a MongoDB controller/client that uses the MongoDB NodeJS Driver.
3. A private MariaDB database and/or the existing TPEN MariaDB database.  This requires a MariaDB controller/client that uses the MariaDB NodeJS Driver.

An abstract create must be able to determine the data source for the create after the fact, or upon additional supplied parameters.

This flow of functionality is required for all CRUD and query.

Bearer token processing and throughput is outside of this scope. 

Adjusting the endpoints to use the database connector is outside of this scope.
> Note that the /manifest directory and endpoint will be adjusted for demonstration purposes only.


database/mongo/controller.mjs
```diff
@@ -0,0 +1,237 @@
...
+            return { "endpoint_error": "find", "status": 500, "message": "There was an error querying the database." }
+        }
+    }
+
+    /**
+     * Insert a document into the database (mongo)
+     * @param data JSON from an HTTP POST request
+     * @return The inserted document JSON or error JSON
+     */
+    async create(data) {
```
cubap, line 142: Let's discuss this...

There are some patterns to consider. We want to return errors without breaking everything and we want to know what class of error it is. 
Inside this `try` we can throw an error and handle it all in a catch that returns

1. an Array like `[err, result]` that can be `[null, result]` for successes for an easy check.
2. an express-style `callback(err, result)` that is controlled by the caller.
3. an Object that always has `{ status, result }` that is easy to check.
4. some combination that does (3) if no callback is provided, for example.
5. just throw the Errors and let the caller handle it.

database/mongo/controller.mjs
```diff
@@ -0,0 +1,237 @@
...
+
+    /**
+     * Update an existing object in the database (mongo)
+     * @param data JSON from an HTTP POST request.  It must contain an id.
+     * @return The inserted document JSON or error JSON
+     */
+    async update(data) {
+        try {
+            //need to determine what collection (projects, groups, userPerferences) this goes into.
+            const data_type = data["@type"] ?? data.type ?? null
```
cubap, line 173: The `?? null` is extra, since you just check for truthiness below.

database/mongo/controller.mjs
```diff
@@ -0,0 +1,237 @@
...
+            let data_id = data["@id"] ?? data._id ?? null
+            let collection = null
+            if (!data_id)
+                return { "endpoint_error": "updateOne", "status": 400, "message": `Cannot find 'type' on this data, and so cannot figure out a collection for it.` }
+            if (!data_type)
+                return { "endpoint_error": "updateOne", "status": 400, "message": `Cannot find 'type' on this data, and so cannot figure out a collection for it.` }
+            collection = discernCollectionFromType(data_type)
```
cubap, line 180: Why not just init `collection` here?

database/mongo/controller.mjs
```diff
@@ -0,0 +1,237 @@
...
+                return { "endpoint_error": "insertOne", "status": 400, "message": `Cannot find 'type' on this data, and so cannot figure out a collection for it.` }
+            const collection = discernCollectionFromType(data_type)
+            if (!collection)
+                return { "endpoint_error": "insertOne", "status": 400, "message": `Cannot figure which collection for object of type '${data_type}'` }
+            const id = this.newID()
+            data["_id"] = id
+            const result = await this.db.collection(collection).insertOne(data)
+            if (result.insertedId) {
+                return data
+            } else {
```
cubap, line 156: These `else`s are removable if you just flip them all. Not wrong, but might be easier to read, since the rest of the function is just kicking away errors and that would leave the default successful behavior as a data return.

database/mongo/controller.mjs
```diff
@@ -0,0 +1,237 @@
...
+        }
+    }
+
+    /**
+     * Update an existing object in the database (mongo)
+     * @param data JSON from an HTTP DELETE request.  It must contain an id.
+     * @return The delete result JSON or error JSON
+     */
+    async remove(id) {
+        return { "endpoint_error": "deleteOne", "status": 501, "message": `Not yet implemented.  Stay tuned.` }
```
cubap, line 206: (aside) these are bad examples for the OSS folks because they are "error-like objects" without actually being Errors.

database/mongo/controller.mjs
```diff
@@ -0,0 +1,237 @@
...
+
+        // We have to query for the object to get the type to know the collection to remove from
+        const data_type = ""
+        let collection = null
+        if (!id)
+            return { "endpoint_error": "deleteOne", "status": 400, "message": `Cannot find 'type' on this data, and so cannot figure out a collection for it.` }
```
cubap, line 212: Should this loop be extracted for reuse? Just washing the input to get the collection and make whatever adjustments are needed to the data?

database/mongo/controller.mjs
```diff
@@ -0,0 +1,237 @@
...
+        } else {
+            return { "endpoint_error": "deleteOne", "status": 500, "message": result.message }
+        }
+    }
+
+    /**
+     * Get by ID.  We need to decide about '@id', 'id', '_id', and http/s 
+     */
+    async getByID(id) {
+        const result = await this.query({ "_id": id })
```
cubap, line 231: It might be useful for this to also accept an `id` and `@id` though I haven't been through the use cases in my mind yet.

database/tiny/controller.mjs
```diff
@@ -0,0 +1,188 @@
+/**
+ * A TinyPEN Controller.  Actions here specifically fetch to the set TinyPEN API URLS.
+ * @see this.URLS
+ * 
+ * @author Bryan Haberberger
+ * https://github.com/thehabes 
+ */ 
+
+import fetch from 'node-fetch'
```
cubap, line 9: Don't import this. Node 18+ already has fetch just built in. This library was a stopgap while it was being done.

database/tiny/controller.mjs
```diff
@@ -0,0 +1,188 @@
...
+        this.URLS.OVERWRITE = process.env.TINYPEN+"overwrite"
+        this.URLS.QUERY = process.env.TINYPEN+"query"
+        this.URLS.DELETE = process.env.TINYPEN+"delete"
+        console.log("TINY API established")
+        console.log(this.URLS)
+    }
+
+    /** Other modules do not connect or close */
+    async connect() {
+        console.log("No need to connect().  The API awaits you!")
```
cubap, line 31: I mean, correct, but we probably don't need a log for this.

database/tiny/controller.mjs
```diff
@@ -0,0 +1,188 @@
...
+     * Use the TinyPEN create endpoint to create the supplied JSON object.
+     * TODO Pass forward the user bearer token from the Interfaced to TinyPEN?
+     * @return the created JSON or Error
+     */
+    async create(data) {
+        return await fetch(this.URLS.CREATE, {
+                method: 'post',
+                body: JSON.stringify(data),
+                headers: {
+                    'Content-Type': 'application/json; charset=utf-8'
```
cubap, line 93: should we call this +ld?

database/tiny/controller.mjs
cubap, line 1: These will all be authenticated and authorized at the route level?

manifest/__tests__/exists_unit.test.mjs
```diff
@@ -10,7 +10,7 @@
 import app from '../../app.mjs'
 
 describe('Manifest endpoint availability unit test (via a check on the app routes). #exists_unit', () => {
-  it('responds to /manifest/id', () => {
+  it('/manifest route is regsitered', () => {
```
cubap, line 13: "registered"

manifest/index.mjs
```diff
@@ -36,7 +36,15 @@ router.use(
...
+// Send a successful response from queryForManifest.  Code is 200, no location header
+export function respondWithManifests(res, manifest){
+   const id = manifest["@id"] ?? manifest.id ?? null
+   res.set("Content-Type", "application/json; charset=utf-8")
+   res.status(200)
+   res.json(manifest)
+}
+
+// Send a successful response with the appropriate JSON.  Code is 200 and has location header
 export function respondWithManifest(res, manifest){
```
cubap, line 48: Are these different?

manifest/manifest.mjs
```diff
@@ -5,51 +5,54 @@
  * https://github.com/thehabes 
  * 
  * */
-
+import fetch from 'node-fetch'
```
cubap, line 8: no thanks

package.json
```diff
@@ -35,12 +36,17 @@
       "express": "^4.18.2",
       "express-urlrewrite": "~2.0.0",
       "http-errors": "~2.0.0",
+      "jest-cli": "^29.7.0",
+      "mariadb": "^3.2.3",
+      "mongodb": "^6.4.0",
       "morgan": "^1.10.0",
-      "nock": "~13.4.0"
+      "nock": "~13.4.0",
+      "node-fetch": "^3.3.2"
```
cubap, line 44: !

cubap (commented):
Nothing frightening, but plenty of considerations. I can work on some of it, if you'd like to assign me in the conversation.

cubap:
I also can't "review" this one, since I _technically_ own it.

database/driver.mjs
```diff
@@ -85,10 +85,8 @@ class dbDriver {
      * @param data JSON from an HTTP POST request
      * @return The inserted document JSON or error JSON
      */
-    async create(data) {
-        const result = await this.controller.create(data)
-        //if (result.endpoint_error) console.error(result)
-        return result
+    async save(data) {
+        return await this.controller.create(data).catch(err => {return err})
```
cubap: This is the same as `return this.controller.create(data).catch(err => err)`

database/tiny/controller.mjs
thehabes: Ah this broke it.  TinyPEN does not understand ld+json.  I'll make an issue.

database/tiny/controller.mjs
thehabes: I believe so.  auth did not exist when this PR started but it does now, I think it is being added at the route level.

cubap (commented):
I support this

PR #68, opened 2024-03-13
auth added token verification and route protection

jsvoo:
I created two functions in the verifyToken middleware(directory). the first is a custom JWT decoder while the other leverages Auth0 for toekn verification or decoding. Both functions are called in the app.mjs to test how they interract with routes. In the end, Only one of these functions would be kept.

### DIFF
--- a/app.mjs
+++ b/app.mjs
@@ -1,68 +1,76 @@
 #!/usr/bin/env node
 
-/** Server initializer for the app.  Registers all the route paths. */ 
+/** Server initializer for the app.  Registers all the route paths. */
 
-import createError from 'http-errors'
-import express from 'express'
-import path from 'path'
-import { fileURLToPath } from 'url'
+import createError from "http-errors"
+import express from "express"
+import path from "path"
+import { fileURLToPath } from "url"
 
 const __filename = fileURLToPath(import.meta.url)
 const __dirname = path.dirname(__filename)
 
-import cookieParser from 'cookie-parser'
-import dotenv from 'dotenv'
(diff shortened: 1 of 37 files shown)


middlewares/verifyToken.mjs
```diff
@@ -0,0 +1,35 @@
...
+export function jwtMiddleware() {
+  const secret = process.env.JWT_SECRET
+
+  return (req, res, next) => {
+    const encryptedToken = req.headers.authorization
+    if (encryptedToken) {
+      try {
+        decodedToken = jwt.verify(encryptedToken, secret)
+        // if (decodedToken.user_id && decodedToken.role){
+        //  req.user = decodedToken // return decodedToken
```
cubap, line 14: This is nice for passing the user along reliably for processing downstream

app.mjs
```diff
@@ -1,68 +1,76 @@
...
+
+const verifyToken = verifyWithAuth0()
+
+
+// app.use(verifyToken) //check all routes
+app.use("/page/*", verifyToken) //check specific routes
+
+app.use("/", indexRouter)
+app.use("/manifest", manifestRouter)
+app.use("/project", projectRouter)
```
cubap, line 66: You can also be like `app.use("/project", verifyToken, projectRouter)` or for only some routes inside of projectRouter

jsvoo:
image

Note: before testing, redefine the process.env.AUDIENCE and pass value or manually add audience in the function as this test has a had time reading from env directly. I reset the audience to read env after testing.

cubap (approved):

Issue #70, opened 2024-03-18
simple getUserProjects() workflow


cubap:
Derived from #69 (see for details), we need the following functionality:

## services endpoint at `GET /projects?`

* Requires authentication and uses the User Agent in the token
* returns a JSON Array of Projects like `[ { id: "123abc", title: "My Project"} ]` 
* surfaces errors from database connector, rejects malformed requests
* leaves a clear space for parameterized requests to filter the response further.

## simple ProjectFactory

* create Project JSON objects for aggregating in the response
* base on the default `{ id, title }` above but allow for the more complex parameters in #69 

## simple ProjectListFactory

* find collections of Projects based on options (just match User as member now)
* Make no changes to database, but lookup by the Agent passed in
* Allow for necessary filters from #69

Issue #69, opened 2024-03-18
getUserProjects(options)


cubap:
Implement an authenticated endpoint (GET /projects) for a logged in user to get their own list of projects. The endpoint will get the agent from the bearer token and use that id as the lookup, returning an unexpanded list of projects, offering only (at this time) project ID and Title.

This is a GET, but for more sophisticated uses, a query string can manipulate the default options:

parameter | default | description
---|---|---
isPublic | undefined | true to include only public projects; false to exclude public projects. (TODO: open GET /publicProjects)
hasRoles[] | ALL | limit based on the group membership for projects. If used multiple times, matching ANY includes it in the return
exceptRoles[] | NONE | remove projects from the returned list based on group membership. Can be used multiple times. Overrules inclusion from `hasRoles[]`
createdBefore | NOW | only include projects with a creation date not before the UNIX seconds value. Combine with `createdAfter` for a single range.
modifiedBefore | NOW | only include projects modified (by anyone) before the UNIX seconds value. Combine with `modifiedAfter` for a single range. Overruled by `created-x` filters.
createdAfter | 0 | inverts `createdBefore`
modifiedAfter | 0 | inverts `modifiedBefore`
hasCollaborators | undefined | true for group sizes >1; false for group sizes >1
tags[] | undefined | may be multiple to match any exact tag (case insensitive) on the project.
fields[] | [id, title] | The returned array will include these listed fields on each entry, adding a value of `undefined` if the property is not present.
count | false | Only return an integer of the length of the list that would have been returned with all other filters applied. Invalidates `fields[]`

## Interfaces

[Many Interfaces will use a version of this](https://github.com/CenterForDigitalHumanities/TPEN-interfaces/issues/3) and it can get expensive to call, so leave room for the caching of this list in TPEN Static. In this way, starting a new project would push up to that list and may avoid making the call into MongoDB. However, any custom filter requests would need to be applied. If we create the need for many vanilla options, we can include more cached options.

## Notes

* These are authenticated so the long and messy GETs are not a big deal and make logging more helpful.
* The "[]" on some parameters is not a standard, but a widely used convention that will likely work well with the NodeJS/Express used now and whatever may be implemented in the future for unambiguous unordered parameter arrays.
* This method and the shape of a Project may not coevolve smoothly, so missing parameters and close spellings may need to be handled on a case-by-case basis. In general, we shouldn't get Errors from this except for malformed URLs, Unauthenticated.
[...]

PR #73, opened 2024-03-25
WIP on Issue 70 simple getuserprojects workflow

BhanuKandula5030:
**Changes Made :** 

**index.mjs:**

- Updated the file to import the dbController from the correct path.
- Defined a new function getUserProjects to fetch projects based on the user ID using the dbController.
- Modified the /project endpoint to use authenticateUser middleware for token validation.
- Integrated the getUserProjects function to fetch user-specific projects.
- Updated response handling to format projects as required and handle errors appropriately.

**line.mjs:**

- Defined the getUserProjects to fetch the  projects based on user ID
- Included dbController to fetch the projects 
- Modified the findLineById function to include a delay to simulate a database operation.
- Updated the function to return line objects based on the provided ID


project/__tests__/end_to_end_unit.test.mjs
```diff
@@ -2,174 +2,36 @@ import projectRouter from '../index.mjs'
...
 import request from 'supertest'
 
-const routeTester = new express()
+const routeTester = express()
 routeTester.use("/project", projectRouter)
 
 describe('Project endpoint end to end unit test (spinning up the endpoint and using it). #end2end_unit', () => {
 
-  it('POST instead of GET.  That status should be 405 with a message.', async () => {
+  it('GET instead of POST to /project/:id. The status should be 405 with a message.', async () => {
```
cubap, line 10: This change makes it incorrect.

project/__tests__/end_to_end_unit.test.mjs
cubap, line 1: so, this eliminates a lot of tests without any clear reason or replacement.

project/index.mjs
cubap, line 1: be careful just taking in someone else's changes. You are removing a lot of code and things that set headers that are useful.

project/project.mjs
```diff
@@ -1,28 +1,54 @@
-/** Logic for the /project endpoint */
-import * as utils from '../utilities/shared.mjs'
-import * as fs from 'fs'
-
-export async function findTheProjectByID(id = null) {
```
cubap, line 5: Why are you removing these?

cubap (changes requested):
This looks like you imported Sandeep's code and bodged your own over the top. There seems to be a lot that is removed that is out of scope for this issue. Am I misunderstanding what you are trying to do?

project/__tests__/end_to_end_unit.test.mjs
BhanuKandula5030: I have corrected in the updated code .I agree with this

project/__tests__/end_to_end_unit.test.mjs
BhanuKandula5030: I have replaced all the tests and now it were passing along with the my code .

project/index.mjs
BhanuKandula5030: I have accommodated the old code and updated the code

project/project.mjs
BhanuKandula5030: I have accommodated the old code and updated the code with my requirements from the issue

BhanuKandula5030:
My issue and Sandeep issue few requirements are similar  in my code but removal of previous code has caused this and i had replaced the old code and updated the code with the issue requirements and included a end to end test to check and all the tests looks good now

project/project.mjs
```diff
@@ -1,12 +1,12 @@
-/** Logic for the /project endpoint */
 import * as utils from '../utilities/shared.mjs'
+import dbDriver from '../database/driver.mjs'
```
thehabes, line 2: Remove unused import

project/project.mjs
```diff
@@ -1,12 +1,12 @@
...
 import * as fs from 'fs'
 
+const database = new dbDriver()
```
thehabes, line 5: This is never used and can be removed

project/project.mjs
```diff
@@ -19,10 +19,58 @@ export async function findTheProjectByID(id = null) {
...
   }
-
-  // Mock the scenario where it takes a couple seconds to look for but not find the Project.
   if (project === null) {
     project = await mockPause
   }
   return project
 }
+
+export async function getUserProjects(userAgent, options) {
```
thehabes, line 28: Add documentation to this function please

project/project.mjs
```diff
@@ -19,10 +19,58 @@ export async function findTheProjectByID(id = null) {
...
+  try {
+    let projectList = [{ id: "123", title: "MyProject" }]
+
+    return projectList
+  } catch (error) {
+    return { status: 500, message: 'Error retrieving user projects from the database', error: error.message }
+  }
+}
+
+export async function responseMapping(projects) {
```
thehabes, line 38: Add documentation to this function please

project/index.mjs
```diff
@@ -159,40 +137,85 @@ export function respondWithProject(req, res, project) {
...
-      utils.respondWithError(res, 500, 'The TPEN3 server encountered an internal error.')
+  } catch (err) {
+    utils.respondWithError(res, 500, 'The TPEN3 server encountered an internal error.')
+  }
+})
+
+// GET /projects route
+router.get('/', async (req, res, next) => {
+  // Check if a project ID is provided
+  if (!req.query.id) {
```
thehabes, line 165: Any /project`/stuff1here` falls into the route above.  No need for this check.   In fact, I can't actually get to the code below because of this check.  I performed this GET request and handed in the user, but it told me I needed to provide an ID.  This handler is for `/` and so will never receive an ID.  This is not functioning as expected.  How can I elicit `getUserProjects()` via a request to the `/project/` endpoint?  You have to have some way to hit http://localhost:3009/project/ and receive a project, that is a minimum requirement for this task and right now it is not possible.

Please fix this so it is possible to get to and test the code after this.

project/index.mjs
```diff
@@ -159,40 +137,85 @@ export function respondWithProject(req, res, project) {
...
+    utils.respondWithError(res, 400, 'Improper request. There was no project ID.')
+    return
+  }
+
+  const { hasRoles, exceptRoles, createdBefore, modifiedBefore, count, fields } = req.query
+
+  try {
+    let projects = await logic.getUserProjects(req.user, req.query)
```
thehabes, line 173: `req.user` is always undefined.  That is not how you detect the user.  To detect the user, you will use the Auth0 middleware to parse the user from the Bearer token in the Authorization header.  As written, this will only ever result in an error, which means you won't ever be able to return a project.  You will need to do something like...
```
Get the token from the Authorization Header
Use auth0 middleware to validate the token and extract the user from the token, or error
Pass the user into logic.getUserProjects()
Respond with the projects found by logic.getUserProjects()
```

So instead of 
```
 let projects = await logic.getUserProjects(req.user, req.query)
```

You need a functioning mock...
```
const mockUserFromTokenProcessing = {"id":"123", "type":"User", "other":"props"}
let projects = await logic.getUserProjects(mockUserFromTokenProcessing , req.query)
```
... that returns a project or multiple project objects.  

You can decide whether or not to use the actual existing /auth/ logic to parse the token and get the user (not recommended for this current scope).  You are not required to use it in this PR.  What is required is a functioning endpoint that returns user projects.

thehabes (changes requested):
Please apply the changes which make this functional and possible to test manually.  I'll look at the tests once I can confirm the route accomplishes the minimal task of returning "user projects" when there is an Authenticated GET request to `/project/`.

project/index.mjs
thehabes: image

thehabes: image

project/project.mjs
BhanuKandula5030: Thanks I have removed it and updated in the latest code .

project/project.mjs
BhanuKandula5030: Thanks I have removed it and updated in the latest code .

project/project.mjs
BhanuKandula5030: I have added the structured documentation for this function.

project/project.mjs
BhanuKandula5030: I have added the structured documentation for this function.

project/index.mjs
BhanuKandula5030: I have fixed the project end point and added the test case to check and it was passing and attaching the postman success response for your reference.

BhanuKandula5030: project end point

project/index.mjs
BhanuKandula5030: I have checked  using the auth but it was throwing : AssertionError [ERR_ASSERTION]: An 'audience' is required to validate the 'aud' claim.So I preferrred not to use the auth logic for this PR .Where the req.user is removed and used the functioning mock as suggested

thehabes (changes requested):
This did not pass manual testing.  There is no longer a /projects endpoint by which to get the users projects.
image

It looks like now you tried to add this to the /project endpoint.  This task requires that there be a new endpoint /projects.  Issue #69 says

> Implement an authenticated endpoint (GET /projects) for a logged in user to get their own list of projects.

**NOTE**
I see this in #70 

image
^
I will be updating this to GET /projects for clarity.

cubap (approved):

auth/__tests__/auth_unit_test.mjs
```diff
@@ -9,7 +9,7 @@ const app = express()
 
 app.use(auth0Middleware())
 
-describe("auth0Middleware #auth_test", () => {
+describe.skip("auth0Middleware #auth_test", () => {
```
cubap, line 12: I hate this

projects/__tests__/end_to_end_unit.test.mjs
```diff
@@ -0,0 +1,281 @@
+import projectsRouter from '../index.mjs'
+import express from 'express'
+import request from 'supertest'
+
+process.env.AUDIENCE = "provide audience to test"
+
+const routeTester = new express()
+routeTester.use("/projects", projectsRouter)
+
+describe.skip('Projects authenticated endpoints end to end unit test (spinning up the endpoint and using it). #end2end_unit', () => {
```
cubap, line 10: I hate this

projects/__tests__/end_to_end_unit.test.mjs
```diff
@@ -0,0 +1,281 @@
...
+      .auth(process.env.TEST_TOKEN, {type: "bearer"})
+    expect(res.statusCode).toBe(200)
+    expect(res.body).toBeTruthy()
+    expect(parseInt(res.text)).not.toBe(NaN)
+  })
+   
+  it('Authenticated GET to /projects?createdBefore=1000&count=true. The status should be 200 with 0 in the body.', async () => {
+    const res = await request(routeTester)
+      .get('/projects?createdBefore=1000&count=true')
+      .auth(process.env.TEST_TOKEN, {type: "bearer"})
```
cubap, line 236: This env is not set anywhere

projects/__tests__/end_to_end_unit.test.mjs
```diff
@@ -0,0 +1,281 @@
...
+    expect(res.statusCode).toBe(200)
+    expect(res.body).toBeTruthy()
+    expect(parseInt(res.text)).toBe(0)
+  })
+   
+  it('Authenticated GET to /projects?modifiedBefore=1000&count=true. The status should be 200 with 0 in the body.', async () => {
```
cubap, line 242: this won't work with the demo test

cubap (approved):

PR #75, opened 2024-03-25
WIP on Issue #69

leandrumartin:
Started implementing /projects endpoint. As of now, any requests to the endpoint return an error, but I have started implementing the basic logic to build the endpoint off of.


projects/__tests__/end_to_end_unit.test.mjs
```diff
@@ -0,0 +1,68 @@
...
+    const res = await request(routeTester)
+      .get('/projects/')
+    expect(res.statusCode).toBe(401)
+    expect(res.body).toBeTruthy()
+  })
+
+  // it('Authenticated GET request with no queries.  The status should be 200.', async () => {
+  //   const res = await request(routeTester)
+  //     .get('/projects/')
+  //     .auth(process.env.TEST_TOKEN, {type: "bearer"})
```
cubap, line 59: hmmm.

projects/index.mjs
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Authenticated endpoint to return a list of a user's projects.
+ * 
+ * @param options A JSON object of options for the project list lookup.
+ * @return An unexpanded list of the user's projects. Returns empty list if none found.
+ */
+export async function respondWithProjects(user, options, res){
+  // Set option defaults
+  let hasRoles       = options.hasRoles       ?? 'ALL'
```
cubap, line 52: If I'm reading this right, these `let`s are not needed and should be `const`. In fact, the one's that change (`NOW` to `Date.now()`) should be preset and then the defaults can be left alone.

projects/index.mjs
```diff
@@ -0,0 +1,105 @@
...
+  if (exceptRoles !== 'NONE') {
+    projects = projects.filter(project => {
+      !project.roles.some(role => exceptRoles.includes(role))
+    })
+  }
+
+  if (createdBefore === 'NOW') {
+    createdBefore = Date.now()
+  }
+  projects = projects.filter(project => createdAfter < project.created && project.crated < createdBefore)
```
cubap, line 78: typo! `&& project.crated`

projects/index.mjs
```diff
@@ -0,0 +1,105 @@
...
+
+  if (modifiedBefore === 'NOW') {
+    modifiedBefore = Date.now()
+  }
+  projects = projects.filter(project => modifiedAfter < project.lastModified && project.lastModified < modifiedAfter)
+
+  if (count) {
```
cubap, line 85: For sematics, I would prefer that if there is a count, return something that looks like a count. Having two `res.send()` that send different types is clearer and easier to manage. Here is the chance to change headers for the type of response as well. `send()` just pushes back the body as text (number here) and `json()` sends back the array as JSON.

projects/projects.mjs
```diff
@@ -0,0 +1,26 @@
+/** Logic for the /projects endpoint */
+import * as fs from 'fs'
+
+export async function getUserProjects(user) {
```
cubap, line 4: okay. I can accept this for now.

cubap (changes requested):
Some minor points of information or correction, but this is generally very good for this scope

projects/__tests__/end_to_end_unit.test.mjs
```diff
@@ -0,0 +1,68 @@
+import projectsRouter from '../index.mjs'
+import express from 'express'
+import request from 'supertest'
+import 'dotenv/config'
```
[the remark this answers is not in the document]
leandrumartin: This is what I had found online, and it seemed to work correctly. I'll change it to the more standard version anyway, though.

projects/index.mjs
leandrumartin: The idea I was going for is that the only valid inputs for this parameter are either "NOW" or a date, and any other input would return an error. I thought this would make it easier to validate the queries, since I would only have to do one check for "NOW" in the same place I check for a valid date.

leandrumartin: My thought process, at least for the createdBefore and modifiedBefore parameters, is that the valid inputs would be either "NOW" or a numerical date. Even if I make Date.now() the default here, I'd have to ensure that if the user manually sends "NOW" then that would be accepted. That would require either changing the createdBefore variable to Date.now(), or creating an entirely new variable to use when actually filtering through the projects[] array. Using `let` seemed cleaner. I'll change the others, though, since those shouldn't change.

thehabes (changes requested):
This code has breaking errors.

When I make a properly authenticated request the app errors out and kills the node process.  This cannot happen.

image

projects/index.mjs
```diff
@@ -0,0 +1,154 @@
+/** Route handler for the /projects endpoint */
+import express from 'express'
+import * as logic from './projects.mjs'
+import * as utils from '../utilities/shared.mjs'
+import cors from 'cors'
+import auth0Middleware from '../auth/index.mjs'
+import 'dotenv/config'
```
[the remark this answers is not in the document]
leandrumartin: Without this import, I get an error from the authentication middleware where Auth0 requires an audience, which is read from my .env file: `AssertionError [ERR_ASSERTION]: An 'audience' is required to validate the 'aud' claim`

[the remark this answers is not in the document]
thehabes: Ah that's right, you want to use the actual auth.  Yes then keep the import.

thehabes (changes requested):
This did not pass manual testing.  There is no way for me to get anything besides an empty array from /projects.

image

The mock projects do not make it back out to me.  However, I know the endpoint is trying to return them.

image

image

A successful /projects request will either give a RESTful error code/message, or return your mock projects

I expected this request to give errors on `hasRoles`, `isPublic`, and `count` but instead only gives errors for `count`.  Maybe this is out of scope for this PR
image

cubap (approved):

cubap:
Just trying to unbury us all here. The code does not connect to the db right now so any authenticated user returns the stubbed projects, which also means there is not full examples for all the parameters possible.

Issue #93, opened 2024-04-04
Connect /project with database(s)


thehabes:
Determine which databases the Project logic will use in order to CRUD Projects and make the connection to those databases.  Generate or change logic for create - update - delete - query actions against those databases.  Update tests.

cubap:
Tests should confirm validation checks for new Projects, mock the driver actions, and respond correctly to expected errors or invalid inputs.

Issue #101, opened 2024-04-08
`isValid(id)` for db controller


cubap:
Specific to the various drivers, we should have access to a method to check for a valid id for lookup, as with `mongoose.Types.ObjectId.isValid('5c0a794449d898a0f4911426')` in mongoose/MongoDB or maybe something like `isValidId = (id) => ObjectId(id).toString() === id` with better try-catch handling for MongoDB. The task here is to implement in the driver and then abstract for the methods.

cubap:
![image](https://github.com/CenterForDigitalHumanities/TPEN-services/assets/1119165/5cb006a3-4c53-40b3-92c4-0b6df998a6b8)

cubap:
@thehabes It works on my machine. All green on the PR for the hotfix. I'll wait to put anything else in until this is pushed through

PR #103, opened 2024-04-15
WIP on Issue #93

leandrumartin:
Added /project/create route to add Projects to local Mongo database, and implemented the basic logic for it. I also added a test for that logic (test with `npm run unitTests`.

### DIFF
--- a/project/index.mjs
+++ b/project/index.mjs
@@ -162,4 +162,21 @@ router.all('/', (req, res, next) => {
   utils.respondWithError(res, 405, 'Improper request method, please use GET.')
 })
 
+router.route('/create')
+  .post(async (req, res, next) => {
+    if (!utils.isValidJSON(res.body)) {
+      utils.respondWithError(res, 400, "Improperly formatted JSON")
+      return
+    }
+    const logicResult = logic.saveProject(res.body)
+    if (logicResult["_id"]) {
+      res.status(201).json(logicResult)
+    } else {
+      utils.respondWithError(res, logicResult.status, logicResult.message)
+    }
+  })
+  .all((req, res, next) => {
+    utils.respondWithError(res, 405, "Improper request method, please use POST.")
+  })
+
 export default router
--- a/project/__tests__/functionality_unit.test.mjs
+++ b/project/__tests__/functionality_unit.test.mjs
@@ -1,5 +1,7 @@
-import {findTheProjectByID} from '../project.mjs'
+import * as logic from '../project.mjs'
 import {validateID} from '../../utilities/shared.mjs'
+
+let test_project = { "type": "Project", "title": "test project" }
 
 // These test the pieces of functionality in the route that make it work.
 describe('Project endpoint functionality unit test (just testing helper functions). #functions_unit', () => {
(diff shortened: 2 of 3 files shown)


project/index.mjs
```diff
@@ -138,6 +138,94 @@ export function respondWithProject(req, res, project) {
...
+    - Check for manifest, create one in the database if it doesn't exist
+  */
+
+  // TEMPORARY stubbing out required keys we need to generate
+  project.creator = "user"
+  project.group = "group"
+  project.slug = "slug"
+
+  // Required keys
+  if (project.created) {
```
thehabes, line 162: Yes something like the Project class could have a function to validate the minimum requirements.

project/index.mjs
```diff
@@ -138,6 +138,94 @@ export function respondWithProject(req, res, project) {
...
+      utils.respondWithError(res, 400, 'Project key "tags" must be an array')
+      return
+    }
+    if (!project.tags.every(tag => typeof tag === 'string')) {
+      utils.respondWithError(res, 400, 'Project key "tags" must be an array of strings')
+      return
+    }
+  }
+
+  req.body["@type"] = "Project"
```
thehabes, line 206: I think this would be a part of the helper function that checks for the minimum requirements.  Sure, we can assume this is what they meant.  However if we do this and they post to this with a `"@type":"User"` it would probably be better to give them a RESTful message about how this is all wrong.

thehabes (commented):
Confirmed that `npm run allTests` passes, and noticed some tests were added to support the new endpoint pattern (I'll look at those closely on an official review).

Confirmed expected "create" functionality.
image

However, I am unable to GET that created project.  @Sandeep-1-Kumar is running into the same issue.
image

Otherwise this looks good, just noting that trouble about being able to GET a project.  Within your scope is the ability to GET a project that was created.

leandrumartin:
@thehabes The GET route is requiring that the project ID be a number, but the IDs autogenerated by Mongo contain letters. Should we simply remove the number requirement?

thehabes:
@leandrumartin 
Correct.  The "id" you are expecting is [conformant with MongoDB](https://github.com/CenterForDigitalHumanities/TPEN-services/issues/108).  This means you are no longer expecting an integer.  You are expecting a "hexidecimal string" aka "string" that contains letters and numbers.  

Outside of your scope, we are working on a [shared isValid() for this new id pattern](https://github.com/CenterForDigitalHumanities/TPEN-services/issues/101).  We might merge that in today.  For now, all you can assume about how to validate yours is that "it is a string".  You can either remove the validation calls in your code, or replace the logic of you "isValid()" with `return true`

project/index.mjs
leandrumartin: Should I put this functionality into a Project class then instead of in the project/index.mjs file where it is now?

cubap: We will want a ProjectFactory that will prepare a database record. This will give us the ability to throw an invalid error with details, but also to assert lots of defaults. For example, `req.user.agent` will be the `creator`, `created: now` can be automatic, if there is no `title` a placeholder can be applied, a Group and OWNER can be applied, etc. It's way more complex than just "is this the minimum?" so let's leave it boxed up for the moment.

PR #106, opened 2024-04-15
User class

jsvoo:
Added user class, tests, and object modifying utilities.
Some of the things I need feedback on include

- db selection: should this happen on instance of user class or is this constant. it's currently set to allow db settings on instance but defaulted to "mongo"
- standard `id`: id || _id ?
- Error handling: most errors have been handled on the dbDriver, I didnt think it was necessary to intercept responses or errors any further

### DIFF
--- a/classes/User/User.mjs
+++ b/classes/User/User.mjs
@@ -87,3 +87,5 @@ const publicDetails = await userObject.getUserById()
 const userProjects = await userObject.getProjects()
 
 console.log(userObject)
+
+
--- a/classes/User/User.mjs
+++ b/classes/User/User.mjs
@@ -1,0 +1,100 @@
+import {removeProperties} from "../../utilities/removeProperties.mjs"
+
+export class User {
+  constructor(database) {
+    this.database = database
+  }
+
+  async getUserById(userId, privateDetails=["profile", "@type", "agent"]) {
+    let user = await this.database.find({
+      _id: userId,
+      "@type": "UserPreferences"
+    })
+    if (Array.isArray(user)) {
+      user = user[0]
+    }
+    return removeProperties(user, ...privateDetails)
+  }
+
+  async getSelf(userId) {
+    // returns full user object, only use this when the user is unthenticated i.e, logged in and getting himself.
+    const user = await this.database.find({
+      _id: userId,
+      "@type": "UserPreferences"
+    })
+    if (Array.isArray(user)) {
+      return user[0]
+    } else {
+      return user
+    }
+  }
+
+  async updateUser(data) {
+    if(!data) return 
+    const previousUser = await this.getSelf(data?._id);
+    const updatedUser = { ...previousUser, ...data }; 
+     const result = await this.database.update(updatedUser);
+     return result
+  }
+
+  async getProjects(userId) {  
+    const projects = await this.database.find({
+      "@type": "Project",
+      author: userId
+    })
+    return projects 
+}
+
+ /**
+  * 
+  * @param projectData JSON object containing project detail including {"author":"user._id"}
+  * @returns the same object if creation is successful
+  */
+  async createProject(projectData) {
+     const newProject = await this.database.save({...projectData, "@type":"Project"})
+    return newProject
+ }
+
+
+
+
+
+
+
+
+async createUser(userData) {
+    const newUser = await this.database.save(userData)
+    return newUser
+  }
+
+
+async deleteUser(userId) {
+    try {
+      await this.database.delete(id, userId)
+      return true
+    } catch (error) {
+      throw new Error("Failed to delete user")
+    }
+  }
+ 
+
+  // Method to update a project
+  async updateProject(data) {
+    // data = {...otherProperties, id:projectId}
+    try {
+      const updatedProject = await this.database.update(data)
+      return updatedProject
+    } catch (error) {
+      throw new Error("Failed to update project")
+    }
+  }
+
+  async deleteProject(projectId) {
+    try {
+      await this.database.deleteProject(projectId)
+      return true
+    } catch (error) {
+      throw new Error("Failed to delete project")
+    }
+  }
+}
--- a/classes/User/User.mjs
+++ b/classes/User/User.mjs
@@ -26,7 +26,7 @@ export class User {
     })
     user = user[0]
     this.id = user?._id
-    const publicUser = includeOnly(user, "profile")
+    const publicUser = includeOnly(user, "profile", "_id")
     this.user = publicUser
     return publicUser
   }
@@ -50,8 +50,7 @@ export class User {
     return updatedUser
   }
 
-  async getProjects() {
-
+  async getProjects() { 
     // this assumes that the project object includes the following properties
     // {
     //   "@type":"Project"
@@ -64,9 +63,7 @@ export class User {
     if(!user) return []
     const allProjects = await database.find({
       "@type": "Project"
-    })
-   
-
+    }) 
     const userProjects = []
     allProjects?.map((project) => {
       if (project.creator === this.id) {
--- a/classes/User/User.mjs
+++ b/classes/User/User.mjs
@@ -1,82 +1,92 @@
-import {removeProperties} from "../../utilities/removeProperties.mjs"
+import dbDriver from "../../database/driver.mjs"
+import {
+  includeOnly,
+  removeProperties
+} from "../../utilities/removeProperties.mjs"
 
 export class User {
-  constructor(database) {
-    this.database = database
+  constructor(user = null, database = "mongo") {
+    this.database = new dbDriver(database)
+
+    if (user instanceof Object) {
+      this.id = user._id
+    } else {
+      this.id = user
+    }
+    this._fetchUser()
   }
 
-  async getUserById(userId, privateDetails=["profile", "@type", "agent"]) {
+  async _fetchUser() {
     let user = await this.database.find({
-      _id: userId,
-      "@type": "UserPreferences"
+      _id: this.id,
+      "@type": "users"
     })
     if (Array.isArray(user)) {
       user = user[0]
     }
-    return removeProperties(user, ...privateDetails)
+    this.id = user?._id 
+    return (this.user = user)
   }
 
-  async getSelf(userId) {
+  async getUserById() {
+    // returns user's public info
+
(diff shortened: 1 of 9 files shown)


cubap:
The 'mongo' designation is probably going to be refactored in the future to something like "local" or "private" so it is appropriate to have in here.

cubap:
Ids can be tricky.

`_id` is the mongoDB default and I think for all private data it is sufficient.
`id` and `@id` are used in RERUM and always refer to a full URL for the JSON object. The end of the `@id` is the hexString from the `_id` in the MongoDB RERUM uses, but you can also use a slug to refer to an object and then the hexString is preempted by the `__rerum.slug` String.

classes/User/User.mjs
```diff
@@ -0,0 +1,110 @@
...
+    // if (Array.isArray(user)) {
+    //   return user[0]
+    // } else {
+    //   return user
+    // }
+  }
+
+  async updateRecord(data) {
+    // updates user object. use with PUT or PATCH
+    if (!data) return
```
cubap, line 55: obviously this is auth only

classes/User/User.mjs
```diff
@@ -0,0 +1,110 @@
...
+    const previousUser = this.user
+    const newRecord = {...previousUser, ...data}
+    const updatedUser = await this.database.update(newRecord)
+    return updatedUser
+  }
+
+  async getProjects() {
+    const projects = await this.database.find({
+      "@type": "Project",
+      creator: this.id
```
cubap, line 65: This would be only the ones that they have created. This needs to be able to find all projects with this id in the group. Right now, a group looks like

```json
{
  "_id": {
    "$oid": "65f48c9a4da4e4b8c656d0d2"
  },
  "slug": "exampleGroup",
  "members": [
    {
      "agent": "https://store.rerum.io/v1/id/5bec93c7d5deb589d44ec1a8",
      "roles": [
        "OWNER",
        "LEADER"
      ]
    }
  ],
  "custom_roles": [
    {
      "OWNER": {
        "label": "Owner",
        "permissions": "ALLOW_ALL"
      },
      "LEADER": {
        "label": "Group Leader",
        "permissions": "ALLOW_MODIFY_MEMBERS"
      }
    }
  ],
  "custom_defaults": {},
  "tags": [
    "GoG"
  ]
}
```

but maybe we should be using the `_id` since it is internal and have `id` instead of `agent`. That would allow an id map in `members` instead, which would be an easy lookup in mongodb `{'members.hexString':{"$exists":true}`

just thinking out loud

cubap: agent is easy to get quickly from the `users` collection and could be left embedded even.

classes/User/User.mjs
```diff
@@ -0,0 +1,110 @@
...
+   * @param projectData JSON object containing project detail including {"creator":"id"}
+   * @returns the same object if creation is successful
+   */
+  async createProject(projectData) {
+    const newProject = await this.database.save({
+      ...projectData,
+      "@type": "Project",
+      creator: this.id
+    })
+    return newProject
```
cubap, line 84: This Response would be an HTTP Response, I believe, so we would want a little step here to validate it before saving and extract the Project from the response to return.

cubap: Now that I'm thinking about it... the User is always known through auth, so createProject() doesn't have to be in the User class at all. Project(data) will create one just fine.

classes/User/User.mjs
```diff
@@ -0,0 +1,110 @@
...
+
+  async deleteUser() {
+    await this.database.delete(user)
+  }
+
+  // Method to update a project
+  async updateProject(data) {
+    // data = {...otherProperties, id:projectId}
+    try {
+      const updatedProject = await this.database.update(data)
```
cubap, line 95: Check for auth and for permissions that allow for editing the project

cubap: another place where this is more properly in the Project class.

classes/User/User.mjs
```diff
@@ -0,0 +1,110 @@
...
+      return updatedProject
+    } catch (error) {
+      throw new Error("Failed to update project")
+    }
+  }
+
+  async deleteProject(projectId) {
+    try {
```
cubap, line 103: definitely in Project instead

projects/projects.mjs
```diff
@@ -0,0 +1,41 @@
+/** Logic for the /projects endpoint */
+
+// Mock projects list for now
+// User -1 for failure, 404 for none found, other for success
+class User {
```
cubap, line 5: This is where you can import your User and replace the getProjects

cubap (commented):
Some comments to consider. Repoint to `development` branch so you can see your own changes more clearly.

classes/User/User.mjs
```diff
@@ -0,0 +1,110 @@
...
+    }
+    this._fetchUser()
+  }
+
+  async _fetchUser() {
+    let user = await this.database.find({
+      _id: this.id,
+      "@type": "users"
+    })
+    if (Array.isArray(user)) {
```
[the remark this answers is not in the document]
cubap: https://github.com/CenterForDigitalHumanities/TPEN-services/blob/main/database/mongo/controller.mjs#L255

it exists

classes/User/User.mjs
```diff
@@ -0,0 +1,110 @@
...
+    } else {
+      this.id = user
+    }
+    this._fetchUser()
+  }
+
+  async _fetchUser() {
+    let user = await this.database.find({
+      _id: this.id,
+      "@type": "users"
```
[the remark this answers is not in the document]
thehabes: That @type is incorrect.  Users look like

```
 {
    _id: '661855c589b46da6f3df060c',
    '@type': 'User',
    ...
}
```

If you are looking something up by `_id` then that is the only property you need for `find()`
Also, `database.getById()` is available and hopefully functional.

```
    /**
     * Get by ID.  We need to decide about '@id', 'id', '_id', and http/s 
     */
    async getByID(id) {
        return await this.find({ "_id": id })
    }
```

thehabes (commented):
You will want to get this to a spot where we can test it.

I have added a commit so that the `/user/:id` endpoint tries to use the `User` class to initialize a `User` object for that `_id` and return the json.  I get an error when I try.

image

classes/User/User.mjs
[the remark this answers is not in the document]
jsvoo: I used the dbDriver class in case there was flexibility with db selection, that class has no `getById()` yet. but I'd review this to only use the mongo controller since it's constant

classes/User/User.mjs
jsvoo: Yes.

classes/User/User.mjs
jsvoo: Yes, I agree. I'm taking out all project-related functions fromt he user class except getUserProjects

thehabes (commented):
This is not functional.  I will need something functional, otherwise it is tough to leave comments for code that is there since we know it will need to be changed.

I performed the same test and made a request for http://localhost:3009/user/6616ff681dc7ba396df3392c, but I did not get that user back.

image

Here is the matching `find()` call in my Mongo, which is successful to prove to you it is there.

image

I committed in my attempt to use it in /userProfile/index.mjs

```
const userObject = await new User(id)
if (userObject) {
  respondWithUserProfile(res, userObject.getUserById())
}
```

package-lock.json
```diff
@@ -27,6 +27,7 @@
             "@jest-mock/express": "^2.0.2",
             "jest": "^29.7.0",
             "jest-esm-transformer": "^1.0.0",
+            "nodemon": "^3.1.0",
```
thehabes, line 30: We probably do not want these changes upstream

package.json
```diff
@@ -20,13 +20,15 @@
    "repository": "github:CenterForDigitalHumanities/rerum_server_nodejs",
    "scripts": {
       "start": "node ./bin/tpen3_services.mjs",
+      "dev": "nodemon ./bin/tpen3_services.mjs",
```
thehabes, line 23: We probably do not want these changes upstream

thehabes (commented):
The app does spin up and function.  Here is the answer to my call to http://localhost:3009/user/660f270574de2125032c04d2/projects

image

```json
[
    {
        "_id": "7890dd4cf38eb4e8ca4048a1",
        "@type": "Project",
        "name": "Creator and Member",
        "creator": "660f270574de2125032c04d2",
        "groups": {
            "members": [
                {
                    "agent": "https://store.rerum.io/v1/id/661d43bc490d045b8ba04d38",
                    "_id": "660f270574de2125032c04d2"
                }
            ]
        }
    },
    {
        "_id": "5670dd4cf38eb4e8ca4048c7",
        "@type": "Project",
        "name": "Creator Only",
        "creator": "660f270574de2125032c04d2",
        "groups": {
            "members": [
                {
                    "agent": "https://store.rerum.io/v1/id/771d43bc490d045b8ba04d39",
                    "_id": "770f270574de2125032c04d3"
                }
            ]
        }
    },
    {
        "_id": "6602dfa6e3e8958d31de210d",
        "@type": "Project",
        "name": "Not Creator But Member Via Agent",
        "creator": "660d801652df1c2243d6d936",
        "groups": {
            "members": [
                {
                    "agent": "https://store.rerum.io/v1/id/661d43bc490d045b8ba04d38"
                }
            ]
        }
    },
    {
        "_id": "1234dfa6e3e8958d31de211e",
        "@type": "Project",
        "name": "Not Creator But Member Via _id",
        "creator": "770f270574de2125032c04d3",
        "groups": {
            "members": [
                {
                    "_id": "660f270574de2125032c04d2"
                }
            ]
        }
    }
]
```
A note on RESTfulness.  If we decide to keep this endpoint around, we will have to update its responses.  This is a usage of the endpoint with a user ID that is not in the database where a RESTful response would be 404 - User not found.
image

The /user endpoint returns my entire profile as intended as well.
image

The same RESTful note applies.  If we are going to keep this endpoint, we need to update its responses.  This should be 404 - User Not Found
image

thehabes:
Noting that there are couple failing tests related to the /user endpoint when doing `npm run allTests`

image

thehabes:
This has been merged into the reserveId branch, but we are leaving it open until we are sure everything is merged together correctly.

PR #109, opened 2024-04-22
101 isvalidid for db controller

cubap:


cubap:
midflow. more to come, but this is the hotfix

thehabes (changes requested):
I performed the following test which I assumed would result in an invalid id.

```javascript
const database = new DatabaseDriver("mongo")
if(database.isValidId("6")){
   console.log("VALID ID '6'.  I can save")
   return await database.save(manifestJSON)
}
else{
   console.log("INVALID ID '6'.  I won't save.")
   return false
}
```
Instead, the logs told me 
> VALID ID '6'.  I can save

If '6' is valid then I am wrong.  My gut instinct is that '6' is not a valid ID.

These were detected as expected

> INVALID ID 'potato' I won't save
> INVALID ID 'anything' I won't save.

database/mongo/__tests__/unit.test.mjs
```diff
@@ -117,3 +117,73 @@ describe("Mongo Database Unit Functions. #mongo_unit #db", () => {
     expect(true).toBeTruthy()
   })
 })
+
```
cubap, line 120: Not sure what merge is off, but it looks like the `timeout` stuff is lost here...

cubap (commented):
I'm pushing this in because I am rage.

PR #117, opened 2024-05-04
9 simplify cors

leandrumartin:
Simplifying reused CORS headers across files.


thehabes (changes requested):
I am unable to start the app after pulling down this code and running `npm install`.  Here is the error I get.

image

thehabes (approved):
Nice, I really like this approach.  We weren't quite ready to move _where_ the CORS is included for the routes.  We are waffling on which routes needs which kind of cores, so this leaves everything in place while we are thinking about it but really reduces the lines of code.

Of note, using the cors.json is a good idea and I like that you can just straight up `import` the JSON file.

PR #118, opened 2024-05-08
private user routes

jsvoo:


thehabes (changes requested):
Tests on this branch within your scope are failing.  I will review when tests are passing.

image

userProfile/privateProfile.mjs
```diff
@@ -29,29 +30,26 @@ router.use(
...
-  const {_id} = await req.user 
-
-  if (!_id) return respondWithError(res, 401, "Unauthorized user")
+router.get("/projects", auth0Middleware(), async (req, res) => {
+  const user = await req.user
+  if (!user) return respondWithError(res, 401, "Unauthorized user") 
 
-  const userObj = new User(_id) 
+  const userObj = new User(user._id)
   const userProjects = await userObj.getProjects()
```
thehabes, line 48: You should add a mechanism to these in case userObj.getProjects() fails so that this endpoint can respond RESTfully about what error may have occurred.  Without any handling on this, it could cause an incorrect response or even a code crash.

thehabes (changes requested):
Tested locally against the TPEN Interfaces work and this functions as expected.  Do some extra handling in the privateProfile.mjs route in case userObj.getProjects() has issues.

userProfile/privateProfile.mjs
```diff
@@ -29,29 +30,26 @@ router.use(
   })
 )
 
-router.get("/profile", async (req, res) => {
-  const user = await req.user 
-  if (!user) return respondWithError(res, 401, "Unauthorized user") 
+router.get("/profile", auth0Middleware(), async (req, res) => {
+  const user = await req.user
+  if (!user) return respondWithError(res, 401, "Unauthorized user")
   const userObj = new User(user._id)
```
thehabes, line 36: You should add a mechanism to these in case userObj.getSelf() fails so that this endpoint can respond RESTfully about what error may have occurred. Without any handling on this, it could cause an incorrect response or even a code crash.

classes/User/User.mjs
```diff
@@ -22,51 +22,102 @@ export class User {
...
       "@type": "User"
     })
-    user = user[0]
-    this.id = user?._id
-    const publicUser = includeOnly(user, "profile", "_id")
-    this.user = publicUser
-    return publicUser
+    .then(resp => {
+      if (resp instanceof Error) { 
+        throw resp;
```
thehabes, line 39: superfluous `;`, get rid of this one and the others like it in this file.  I also don't think it will make it here if the resp is an Error so you might not need this if at all (I am not sure 100% sure about this).

classes/User/User.mjs
```diff
@@ -22,51 +22,102 @@ export class User {
...
-    const user = await database.find({
+    // returns full user object, only use this when the user is unauthenticated i.e, logged in and getting himself. 
+    return  database.find({
       _id: this.id,
       "@type": "User"
     })
-    return user[0]
+    .then(resp => {
+      if (resp instanceof Error) { 
+        throw resp;
```
thehabes, line 61: I don't think it will make it here if the resp is an Error so you might not need this if at all (I am not sure 100% sure about this).

classes/User/User.mjs
```diff
@@ -22,51 +22,102 @@ export class User {
...
+    const newRecord = {...previousUser, ...data} 
+
+    return database.update(newRecord).then((resp)=>{
+      console.log("entered here")
+      if(resp instanceof Error){ 
+        throw(resp)
+      }
+     return resp
+      
+    }).catch((error)=> Promise.reject(error))
```
thehabes, line 89: You do this different than the other ones.  In the other ones you do 

```
.catch(err => { 
     throw err
})
```

Unless there is a logical reason to do it this way, stay consistent.

classes/User/User.mjs
```diff
@@ -22,51 +22,102 @@ export class User {
...
+    if (resp instanceof Error) { 
+      throw resp;
+    }  
+    return resp[0];
+  })
+  .catch(err => {  
+    throw err; 
+  });
+
+  // const user = await database.find({agent, "@type":"User"})
```
thehabes, line 115: remove comment junk

classes/User/User.mjs
```diff
@@ -22,51 +22,102 @@ export class User {
...
   
-  return user[0]
+  // return user[0]
```
thehabes, line 117: remove comment junk

classes/User/User.mjs
```diff
@@ -87,29 +138,53 @@ async getByAgent(agent){
...
-          if (member.agent === user?.agent || member._id === this.id) {
-            userProjects.push(project)
-          }
-        })
+    if (!user) {
+      err_out.message = "User account not found"
+      err_out.status = 404 
+      throw err_out
+    }
+    // const allProjects = await database.find({
```
thehabes, line 153: remove comment junk

classes/User/User.mjs
```diff
@@ -87,29 +138,53 @@ async getByAgent(agent){
...
+              userProjects.push(project)
+            }
+          })
+        }
+      })
+  
+      return userProjects
+    }).catch((error)=>{throw error})
+
+
```
thehabes, line 178: Too many newlines

database/mongo/controller.mjs
```diff
@@ -213,6 +213,7 @@ class DatabaseController {
      * @return The inserted document JSON or error JSON
      */
     async update(data) {
+        console.log(data)
```
thehabes, line 216: remove console.log() junk

userProfile/privateProfile.mjs
```diff
@@ -29,29 +30,56 @@ router.use(
   })
 )
 
-router.get("/profile", async (req, res) => {
-  const user = await req.user 
-  if (!user) return respondWithError(res, 401, "Unauthorized user") 
+router.get("/profile", auth0Middleware(), async (req, res) => {
+  const user = await req.user
```
thehabes, line 34: I do no think you need this `await`

userProfile/privateProfile.mjs
```diff
@@ -29,29 +30,56 @@ router.use(
...
+  if (!user) return respondWithError(res, 401, "Unauthorized user")
+  
   const userObj = new User(user._id)
-  const userProfile = await userObj.getSelf()
-  res.set("Content-Type", "application/json; charset=utf-8") 
-
-  res.status(200).json(userProfile)
+  userObj.getSelf()
+  .then(userData => {  
+    res.status(200).json(userData);
```
thehabes, line 40: Remove superfluous `;` and all others like it in this file.

userProfile/privateProfile.mjs
```diff
@@ -29,29 +30,56 @@ router.use(
...
+    res.status(error.status || error.code || 500).json({
+      error: error.message || 'An error occurred while fetching the user data.',
+      status: error.status || 'Error'
+    });
+  })
 })
 
-router.get("/projects", async (req, res) => {
-  const {_id} = await req.user 
+router.get("/test/profile", auth0Middleware(),  (req, res) => {
```
thehabes, line 50: Remove test junk.  This looks like a route you were using for testing something.

userProfile/privateProfile.mjs
```diff
@@ -29,29 +30,56 @@ router.use(
...
+      res.status(error.status || 500).json({
+        error: error.message || 'An error occurred while fetching the user data.',
+        status: error.status || 'Error'
+      }); 
+    })
+  })
 
-  if (!_id) return respondWithError(res, 401, "Unauthorized user")
+router.get("/projects", auth0Middleware(), async (req, res) => {
+  const user = await req.user
```
thehabes, line 69: You need to update this route to do the error throughput like you did with "/profile".

image

thehabes (changes requested):
Still some work to do, see inline comments.

userProfile/privateProfile.mjs
jsvoo: I'm unable to replicate this. The error comes out fine on my end
image


for the following error

image

classes/User/User.mjs
```diff
@@ -1,5 +1,5 @@
 import dbDriver from "../../database/driver.mjs"
-let err_out = Object.assign(new Error(), {"status":500, "message":"N/A"})
+let err_out = Object.assign(new Error(), {status: 500, message: "N/A"})
```
cubap, line 2: I don't think this `status` or `message` are ever used, so this is sotr of just `new Error()`. Maybe it doesn't need to be here at all.

cubap (commented):
This looks good. I added a couple commits. If they look okay, I'm good with a merge.

thehabes (changes requested):
The app did not pass manual testing.

I logged in at users.t-pen.org and get a fresh user token, and added that token to my requests in talend API.  here are the results.

**/my/projects**
image


```
node:_http_server:352
    throw new ERR_HTTP_INVALID_STATUS_CODE(originalStatusCode);
          ^

RangeError [ERR_HTTP_INVALID_STATUS_CODE]: Invalid status code: undefined
    at ServerResponse.writeHead (node:_http_server:352:11)
    at ServerResponse.writeHead (E:\tpen3-services\node_modules\on-headers\index.js:44:26)
    at ServerResponse._implicitHeader (node:_http_server:338:8)
    at write_ (node:_http_outgoing:963:9)
    at ServerResponse.end (node:_http_outgoing:1080:5)
    at ServerResponse.send (E:\tpen3-services\node_modules\express\lib\response.js:233:10)
    at respondWithError (file:///E:/tpen3-services/utilities/shared.mjs:35:23)
    at file:///E:/tpen3-services/userProfile/privateProfile.mjs:64:5
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  code: 'ERR_HTTP_INVALID_STATUS_CODE'
}
```
This errors caused the app to crash for the /my/projects route.

**/my/profile**
image

The app did not crash but it does throw this 500 error.  Make sure these routes are functional and try to handle errors in a way that don't cause app crashes, when possible.

jsvoo:
image
image

thehabes (changes requested):
I am not sure this does what you want.  You are testing the /my/projects and /my/profile endpoints.  The golden rule is that if I do something to break those endpoints, at least one of your tests should catch it.

With the golden rule in mind, I went into app.js and commented out this line of code
```
//app.use('/my',  privateProfileRouter)
```
This effectively removed these endpoints from the app.  This is a breaking change.

image

I ran `npm run userClassTest` to make sure it was caught by the tests.  All tests passed.

image

We need failing tests in this scenario.

cubap:
> 
> We need failing tests in this scenario.

It doesn't have to be a "Class Test" that fails, since this is about an available endpoint, but something in the API tests should catch a missing endpoint. The tests on the class do seem to test the class correctly.

thehabes:
From the perspective of the Class, yes I do think that part is doing OK.  The tests start to fail when I go into /classes/User/User.mjs and remove functions.

jsvoo:
image
image

cubap (approved):

PR #119, opened 2024-05-30
Import projects

jsvoo:


Import/index.mjs
```diff
@@ -0,0 +1,77 @@
...
+       // Process manifest into new Project
+      //Return newProject
+      // UNDER CONSTRUCTION
+
+      let newProject = {}
+
+      const iiifManifest = new Manifest(manifest["@id"]) 
+      const sequences = iiifManifest.getSequences()
+
+      newProject.sequences = sequences
```
thehabes, line 39: These are `Pages` for a TPEN3 Project.  A TPEN3 projects does not have the field "sequences".  The Manifest it produces might.  In the scope of TPEN3, each item in the sequence is a Page.  Each item in the sequence may have an AnnotationPage associated with it, which are the Lines of that Page.  You are processing this Manifest for TPEN3 Project metadata, Layers, Pages and Lines

Import/index.mjs
```diff
@@ -0,0 +1,77 @@
...
+
+  }
+
+  static async importProject(projectId) {
+    return Project.fetchManifest(projectId)
+      .then((manifest) => { 
+        return Project.processManifest(manifest)
+      })
+      .then((newManifest) => {
+        const project = new Project(newManifest)
```
thehabes, line 58: processManifest says `Process manifest into new Project`.

What you have in this `then()` is a new TPEN3 Project, not a Manifest.  Isn't new Project(project) redundant?  Or will processManifest(manifest) return a Manifest?

Remember that Project !== Manifest

A Project has Pages, Layers, Lines that it can assemble and return as a Manifest.

Import/index.mjs
```diff
@@ -0,0 +1,61 @@
+import {Manifest} from "manifesto.js"
+import dbDriver from "../database/driver.mjs"
+
+const database = new dbDriver("mongo")
+
+export class Project {
```
thehabes, line 6: Careful here.  We need a /classes/Project/Project.mjs.

That class will be able to construct and import TPEN3 Projects.

A service route would use the Project.mjs for doing things like `services.t-pen.org/project/create?manifestURI=https://iiif.io/api/cookbook/recipe/0001-mvm-image/manifest.json`

Import/index.mjs
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  static async processManifest(manifestId) {
+    return Project.fetchManifest(manifestId)
+      .then((manifest) => {
+        let newProject = {}
+        newProject.title = manifest.title
+        newProject.label = manifest.label
+        newProject.metadata = manifest.metadata
+        newProject.pages = manifest.items ?? manifest.sequences
```
thehabes, line 33: Exactly right, this is all you can really get from the Manifest, these are each of the pieces.

This is where the TPEN 3 Pages come from.  However, `manifest.items` and `manifest.sequences` do not contain TPEN 3 Pages, they contain Canvases.  You will have to import classes/Page/Page.mjs and be prepared to use that to process Canvases.

So you will need to loop those and make TPEN 3 Pages (which may contain Layers and Lines based off the Canvas AnnotationList or AnnnotationPage).  Then you can associated those TPEN 3 Pages with this TPEN 3 Project.

```javascript
//manifest.items and manifest.sequences have a different structure, so be careful with how you loop it
const canvases = getCanvasesFromManifest(manifest)
newProject.pages = canvases.map(canvas => {
    // createPage will also process the Annotations into Layers/Lines
	return Page.createPage(canvas)
})
```

Import/index.mjs
```diff
@@ -0,0 +1,61 @@
...
+import dbDriver from "../database/driver.mjs"
+
+const database = new dbDriver("mongo")
+
+export class Project {
+  constructor(data) {
+    this.data = data
+  }
+
+  async create(payload) {
```
thehabes, line 11: This belong in the /class/Project/Project.mjs class.

Import/project.mjs
```diff
@@ -0,0 +1,53 @@
...
+      "Link",
+      "X-HTTP-Method-Override"
+    ],
+    exposedHeaders: "*",
+    origin: "*",
+    maxAge: "600"
+  })
+)
+
+router.get("/", async (req, res) => {
```
thehabes, line 30: A note this this is likely actually a part of the project route, like `/project/import?data=uri` or `/project/create?data=uri`

Import/index.mjs
```diff
@@ -0,0 +1,86 @@
...
+        return savedProject
+      })
+      .catch((err) => {
+        console.error("Failed to import project:", err)
+        throw err
+      })
+  }
+}
+
+// ToDo before PR
```
cubap, line 84: these comments can be pulled out of here

Import/index.mjs
cubap, line 1: Should these be deleted now if they are absorbed into Project already?

jsvoo: yes, that's now done

classes/Project/ImportProject.mjs
```diff
@@ -0,0 +1,71 @@
...
+    })
+  }
+
+  static async processManifest(manifest) {
+    let newProject = {}
+    newProject.title = manifest.label
+    newProject.label = manifest.label
+    newProject.metadata = manifest.metadata
+     let canvas;
+    if (manifest.items) {
```
cubap, line 26: A great place for a `?` or `??` but this works

cubap (approved):

thehabes (changes requested):
Tested with https://iiif.io/api/cookbook/recipe/0001-mvm-image/manifest.json

My main issue is that after the Project is created there does not appear to be a way for me to get to the Canvas/Content found in https://iiif.io/api/cookbook/recipe/0001-mvm-image/manifest.json by just knowing the TPEN3 Project.

app.mjs
```diff
@@ -27,6 +27,7 @@ import lineRouter from './line/index.mjs'
 import projectsRouter from './projects/index.mjs'
 import userProfileRouter from './userProfile/index.mjs'
 import privateProfileRouter from './userProfile/privateProfile.mjs'
+import testRouter from './Import/project.mjs'
```
thehabes, line 30: remove this from PR

app.mjs
```diff
@@ -65,6 +66,7 @@ app.use('/page', pageRouter)
 app.use('/projects', projectsRouter)
 app.use('/user', userProfileRouter)
 app.use('/my',  privateProfileRouter)
+app.use('/test',  testRouter)
```
thehabes, line 69: Remove this from PR

classes/Project/ImportProject.mjs
```diff
@@ -0,0 +1,142 @@
...
+      })
+  }
+
+  static async processManifest(manifest) {
+    let newProject = {}
+    newProject.title = manifest.label
+    newProject.label = manifest.label
+    newProject.metadata = manifest.metadata 
+    
+newProject["@context"]= manifest["@context"]
```
thehabes, line 26: duplicated line.  Also, it should be
```
newProject["@context"] = "http://t-pen.org/3/context.json"
newProject.manifest = manifest["@id"] ?? manifest.id
```

Make sure you are capturing the Manifest URI for the project.

classes/Project/ImportProject.mjs
```diff
@@ -0,0 +1,142 @@
...
+        return savedProject
+      })
+      .catch((err) => {
+        console.error("Failed to import project:", err)
+        throw err
+      })
+  }
+}
+
+let one = {
```
thehabes, line 78: remove from PR

classes/Project/ImportProject.mjs
```diff
@@ -0,0 +1,142 @@
...
+  viewer: "https://static.t-pen.org/#65f48b254da4e4b8c656d0d1",
+  license: "CC-BY",
+  manifest: "https://external.example.com/manifest.json",
+  tools: [],
+  options: {},
+  tags: ["gog"],
+  title: "First Project"
+}
+
+let two = {
```
thehabes, line 99: remove from PR

classes/Project/ImportProject.mjs
```diff
@@ -0,0 +1,142 @@
...
+  name: "A good name",
+  layers: ["#LayerId1"],
+  manifest: "https://iiif.io/api/cookbook/recipe/0001-mvm-image/manifest.json",
+  creator: "https://store.rerum.io/v1/id/hash",
+  group: "#GroupId",
+  tools: [],
+  options: {}
+}
+
+let three = {
```
thehabes, line 113: remove from PR

thehabes (changes requested):
Still some clean up.

You still need to change your response.  It should not contain `status`, `message`, and `data`.  It should just be the new project., the part is is `response.data` here.  It still looks like this for me, which will not pass any of our Project validators. 

image

Pay attention to `@context` because the value should be the TPEN 3 context, not the context from the manifest.  You also forgot the capture the `manifest` as part of the project object.

thehabes (approved):

PR #122, opened 2024-07-09
User cleanup

jsvoo:
This PR incluedes the following implementation on the User ecosystem:
- replaced stubs with actual methods
- added /user/:id to return user public profile
- rewrite validateID to use isValid method in DBController
- replaced stub tests with actual tests
- remove redundant methods, files, tests, ...


cubap (approved):
this looks real good

PR #123, opened 2024-07-12
Project cleanup

jsvoo:


classes/Project/Project.mjs
```diff
@@ -1,33 +1,36 @@
...
+  message: "Unknown Server error"
+})
 
 const database = new dbDriver("mongo")
 
 export default class Project {
+  #creator
   constructor(userId) {
-    this.creator = userId
+    this.id = userId
```
cubap, line 15: I'm not quite sure I get this...
A `new Project()` would always need a `userId` to invoke. What if I am a Project contributor and want to instantiate the class to do something like `getByID()` or some team modification? I know JS Classes are not exactly Java ones so this can sort of work, but I'm nervous.

classes/Project/Project.mjs
```diff
@@ -1,33 +1,36 @@
...
-      err_out.message =   validation.errors
+      err_out.message = validation.errors
       throw err_out
     }
 
     try {
-     return await database.save({...payload, "@type": "Project"})
-    
-    } catch (err) { 
+      return await database.save({...payload, "@type": "Project"})
```
cubap, line 32: I would rather the driver handle a `saveProject(payload)` than hardcode the `@type` convenience in here.

classes/Project/Project.mjs
```diff
@@ -45,25 +48,19 @@ export default class Project {
...
     return database
-      .getById(projectId)
-      .then((resp) => {
-        resp.json()
-      })
-      .catch((error) => {
-        console.log(error)
-        throw error
-      })
+      .getById(projectId, "Project")
```
cubap, line 58: this is a pattern I'd be okay with in line 32, perhaps.

classes/Project/Project.mjs
```diff
@@ -45,25 +48,19 @@ export default class Project {
...
+      .then((resp) => (resp?.length ? resp[0] : resp))
```
cubap, line 59: What are we expecting? I think if `getById` gives us more than just one thing, we are doing it wrong.

project/index.mjs
```diff
@@ -267,107 +261,99 @@ async function createNewProject(req, res) {
...
-  .route("/create")
-  .post(async (req, res, next) => {
-    // TODO: Add authentication to this endpoint
-    if (!utils.isValidJSON(req.body)) {
-      utils.respondWithError(res, 400, "Improperly formatted JSON")
-      return
-    }
-    await createNewProject(req, res)
-  })
-  .all((req, res, next) => {
```
cubap, line 285: This `all()` may still be appropriate here. That would change the test above as well, fwiw

project/index.mjs
```diff
@@ -267,107 +261,99 @@ async function createNewProject(req, res) {
...
-    utils.respondWithError(
+    const newProject = await projectObj.create(project)
+    res.status(200).json(newProject)
+  } catch (error) {
+    respondWithError(
       res,
-      500,
-      "The TPEN3 server encountered an internal error."
+      error.status || error.code || 500,
+      error.message || "Unknown server error"
```
cubap, line 286: This might be safer as `??` but I'm not certain

cubap (changes requested):
This is pretty good. We can talk through some of it next week.

classes/Project/Project.mjs
jsvoo: `getById` uses databse find() which returns an array. in this case, it'd always be an array of length 1. hence resp[0]. I put resp as a fail safe in case we rewrite the getById to return an object

project/index.mjs
jsvoo: Pretty much the same thing. it checks in the other in which they appear.

project/index.mjs
```diff
@@ -267,107 +261,118 @@ async function createNewProject(req, res) {
...
+      return respondWithError(
+        res,
+        400,
+        "The TPEN3 project ID provided is invalid"
+      )
+    }
+
+    const projectObj = new Project(id)
+    projectObj
+      .getById(id)
```
cubap, line 351: This is how it works now, but it feels goofy. I think `new Project(id)` should just get that Project and return it to you. Just a comment, here, though, not a change request.

utilities/shared.mjs
```diff
@@ -19,20 +21,25 @@ export function isValidJSON(input=""){
  * @param input A string which should be a valid Integer number
  * @return boolean For whether or not the supplied string was a valid Integer number
  */ 
-export function validateID(id){
-   if(!isNaN(id)){
+export function validateID(id, type="mongo"){
```
cubap, line 24: We will eventually remove this entirely from the utils, since the db drivers will be the authorities on what's a valid id.

cubap (changes requested):
I like this a lot. I have a few comments to consider.

classes/Project/Project.mjs
```diff
@@ -45,25 +50,19 @@ export default class Project {
...
   }
 
   async getProjects() {
```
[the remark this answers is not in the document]
jsvoo: I agree!
I'd take this out

cubap (approved):
I completely support this.

PR #124, opened 2024-07-17
Development

jsvoo:


app.mjs
```diff
@@ -22,9 +22,12 @@ import cors from 'cors'
 import indexRouter from './index.mjs'
 import manifestRouter from './manifest/index.mjs'
 import projectRouter from './project/index.mjs'
+import projectsRouter from './project/projects.mjs'
```
cubap, line 25: our pattern everywhere else is for `index.mjs` in these. I think we should stick with it.

cubap: in some cases, both exist and are redundant with each other

auth/__tests__/auth_unit_test.mjs
```diff
@@ -1,32 +1,43 @@
...
 import auth0Middleware from "../index.mjs"
+import { ObjectId } from "mongodb"
 
 process.env.AUDIENCE = "provide audience to test"
 // this test has a had time reading env directly. add
 
 const app = express()
 
 app.use(auth0Middleware())
+const timeOut = process.env.TEST_TIMEOUT ?? 5000
```
cubap, line 12: These types of constants should probably be CAPS_CASED. Thoughts?

auth/__tests__/auth_unit_test.mjs
```diff
@@ -1,32 +1,43 @@
...
+    "No user should be found on req if token is invalid",
+    async () => {
+      const res = await request(app).get("/protected-route")
+      expect(res.req.user).toBeUndefined()
+    },
+    timeOut
+  )
 
-  it("should set req.user with payload from auth and call next", async () => {
+  it.skip("should set req.user with payload from auth and call next", async () => {
```
cubap, line 34: updated and skipped... is this a test we can reinstate? or remove?

classes/Project/Project.mjs
```diff
@@ -0,0 +1,65 @@
+import dbDriver from "../../database/driver.mjs"
+import {validateProjectPayload} from "../../utilities/validatePayload.mjs"
+import {User} from "../User/User.mjs"
+import ProjectFactory from "./ProjectFactory.mjs"
+
+let err_out = Object.assign(new Error(), {
```
cubap, line 6: Is this ever used as a 500? This might be simpler as just a `new Error()` or removes and declared when needed below.

classes/Project/Project.mjs
```diff
@@ -0,0 +1,65 @@
...
+const database = new dbDriver("mongo")
+
+export default class Project {
+  #creator
+  constructor() {}
+
+  /**
+   * @param {any} userAgent
+   */
+  set creator(userAgent) {
```
cubap, line 20: Moving forward, a `creator` will have two ways of being known: `id` and `agent`. We have cleverly aligned them such that `https://store.rerum.io/v1/id/hexadecimal` has the `_id` "hexadecimal".

So what should we be storing here? What would a developer want in the Project being used for services? Ultimately, the `creator` attached to the saved annotations will be the full agent, but that will be automatically added by the API and never rely on the contributor to add.

classes/Project/Project.mjs
```diff
@@ -0,0 +1,65 @@
...
+    if (!projectId) {
+      err_out.message = "Project ID is required"
+      err_out.status = 400
+      throw err_out
+    }
+
+    return database.remove(projectId)
+  }
+
+  async getById(projectId) {
```
cubap, line 53: We might expect that `const myProject = new Project(id)` is the "get by ID" and this function is not helpful on an empty project since it doesn't modify the instance of the class, but returns a completely new Project.

classes/Project/ProjectFactory.mjs
```diff
@@ -0,0 +1,86 @@
+import {Page} from "../Page/Page.mjs"
+import Project from "./Project.mjs"
+
+let err_out = Object.assign(new Error(), {
```
cubap, line 4: same error comment as above

classes/Project/ProjectFactory.mjs
```diff
@@ -0,0 +1,86 @@
...
+  status: 500,
+  message: "Unknown Server error"
+})
+
+export default class ProjectFactory {
+  constructor(data) {
+    this.data = data
+  }
+
+  static async fetchManifest(url) {
```
cubap, line 14: Other libraries use the name like `loadManifest(url)` which may be preferable in case we start using caching more aggressively or allowing for other valid pipelines.

classes/Project/ProjectFactory.mjs
```diff
@@ -0,0 +1,86 @@
...
+      })
+      .catch((err) => {
+        err_out.status = err.status??404
+        err_out.message = err.message??"Manifest not found. Please check URL" 
+        console.log(err)
+        throw err_out
+      })
+  }
+
+  static async processManifest(manifest) {
```
cubap, line 27: This name should be more descriptive, thinking about how it'll appear in code. 

```javascript
let ProjectForDB = await ProjectFactory.processManifest(manifest)
database.save(ProjectForDB, "Project")
```

Perhaps a name like `databaseObjectFromManifest(manifest)` or `asDBObject(manifest)` would be more logical an appendage to `ProjectFactory`?

classes/Project/ProjectFactory.mjs
```diff
@@ -0,0 +1,86 @@
...
+      err_out.status = 404
+      err_out.message = "No manifest found. Cannot process empty object"
+      throw err_out
+    }
+    let newProject = {}
+    newProject.title = manifest.label
+    newProject.metadata = manifest.metadata
+
+    newProject["@context"] = "http://t-pen.org/3/context.json"
+    newProject.manifest = manifest["@id"] ?? manifest.id
```
cubap, line 38: Not required now, but this is probably where we would want to start considering using manifesto or vault to access these types of things.

classes/User/User.mjs
```diff
@@ -0,0 +1,167 @@
...
+    // database = new DatabaseController() // has a 'find' error that breaks the app at instance. when this this fixed this line will replace the above
+
+    if (user instanceof Object) {
+      this.id = user._id
+    } else {
+      this.id = user
+    }
+  }
+
+  async getUserById() {
```
cubap, line 22: same criticism as `Project.getById()` above

database/driver.mjs
```diff
@@ -116,6 +116,57 @@ class dbDriver {
...
+        return this.controller.getById(id, collection).catch(err => err)
+    }
+    /**
+     * 
+     * @param {*} payload body of project to be saved
+     * @param {*} collection collection of table
+     * @returns 
+     */
+
+    async saveProject(payload, collection) {
```
cubap, line 139: I feel `saveProject()` should assume the Project collection.

database/driver.mjs
```diff
@@ -116,6 +116,57 @@ class dbDriver {
...
+    isValidId(id) {
+        return this.controller.isValidId(id)
+    }
+
+    /**
+     * Create a valid ID from a supplied string or number.
+     * @param id The id to validify.
+     * @return string | number
+     */
+    asValidId(id) {
```
cubap, line 167: maybe this can be combined with `reserveId(seed)`

classes/Project/Project.mjs
jsvoo: I believe the creator on the project is better off being an `agent`. Especially since the `id` can be extracted from the same if needed. Maybe we shouldn't be setting creator for this. we have access to the user's token so we can add the agent from it to the project payload when a project is being created.

cubap: Hash, not URI. At least 2:1. Sorry Victor

cubap: deprecate user.agent to user._id

thehabes: Think about 

https://store.rerum.io/v1/id/62572ba71d974d1311abd673

vs

http://localhost:3009/user/62572ba71d974d1311abd673

We don't want the user-creator to be https://store.rerum.io/v1/id/62572ba71d974d1311abd673 because that is not a representation of the TPEN user.

classes/Project/ProjectFactory.mjs
cubap: `ProjectFactory.DBObjectFromManifest(manifest)`

database/driver.mjs
cubap: remove `asValidId()` wherever it appears for "reserveId()"

thehabes: Suspicious that asValidId is actually unused.
image

classes/Project/ProjectFactory.mjs
thehabes: You could also add a little documentation for this, as it is a slightly complex and rather critical thing to understand.

classes/Project/ProjectFactory.mjs
thehabes: Ugh, definitely.  I will need to learn them too.

auth/index.mjs
```diff
@@ -33,19 +34,38 @@ function auth0Middleware() {
...
+  const agent = payload["http://store.rerum.io/agent"]
+  if(!agent){
+    const err = new Error("Invalid token, missing agent claim")
+    err.status = 401
+    next(err)
+    return
   }
 
+ 
+try {
```
thehabes, line 55: Some goofy indexing going on here.

classes/User/User.mjs
```diff
@@ -0,0 +1,167 @@
+import dbDriver from "../../database/driver.mjs"
+let err_out = Object.assign(new Error(), {status: 500, message: "N/A"})
```
thehabes, line 2: Maybe move this to after the imports.

classes/User/User.mjs
```diff
@@ -0,0 +1,167 @@
...
+import DatabaseController from "../../database/mongo/controller.mjs"
+import {
+  includeOnly,
+  removeProperties
+} from "../../utilities/removeProperties.mjs"
+
+const database = new dbDriver("mongo")
+export class User {
+  constructor(user) {
+    // database = new DatabaseController() // has a 'find' error that breaks the app at instance. when this this fixed this line will replace the above
```
thehabes, line 13: delete this comment, it is no longer relevant.

classes/User/User.mjs
```diff
@@ -0,0 +1,167 @@
...
+    try {
+      const user = await database.save({...data, "@type": "User"})
+      return user
+    } catch (error) {
+      throw error
+    }
+  }
+
+  async getProjects() {
+    // this assumes that the project object includes the following properties
```
thehabes, line 114: This is a lot of comment to have here and would be better as Javadoc style.

database/driver.mjs
thehabes: I can make it even easier on you and tell you that this function should be removed from this file.  It is neve used, and does what save() does.  saveProject() is used elsewhere in this codebase, but not through the driver.

classes/Project/Project.mjs
```diff
@@ -0,0 +1,65 @@
...
+    const validation = validateProjectPayload(payload)
+
+    if (!validation.isValid) {
+      err_out.status = 400
+      err_out.message = validation.errors
+      throw err_out
+    }
+
+    try {
+      return database.saveProject(payload, "Project")
```
thehabes, line 34: This should be `database.save(payload)`

database/driver.mjs
jsvoo: can `reserveId() still replace asValidId() in the contexts above? @cubap @thehabes

PR #127, opened 2024-07-26
126 error fixes

jsvoo:


classes/Project/Project.mjs
```diff
@@ -42,24 +39,62 @@ export default class Project {
...
+      message: "Project data is not loaded.",
+    };
+  } 
 
-  async getById(projectId) {
-    if (!projectId) {
-      err_out.message = "Project ID is required"
-      err_out.status = 400
-      throw err_out
+  if (this.projectData.creator === userAgent) {
```
cubap, line 56: I'm happy to take this in, but we probably would have the agent/user already on the request, since the Project is only returned behind auth.

classes/Project/Project.mjs
```diff
@@ -42,24 +39,62 @@ export default class Project {
...
   }
 
+  return {
+    hasAccess: false,
+    message: "User has no access to this project.",
+  };
+}
+async #getById(projectId) {
+  if (!projectId) {
+    throw {status: 400, message: "Project ID is required"}
```
cubap, line 93: Since this isn't a Request, the status would not be 400. I get the concept, but it isn't strictly right.

cubap: One choice could be to remove the guard clause and just alias database.getById() and handle errors if needed.

userProfile/privateProfile.mjs
```diff
@@ -10,22 +10,23 @@ router.use(
   cors(common_cors)
 )
 
-router.get("/profile", auth0Middleware(), async (req, res) => {
+router.route("/profile").get(auth0Middleware(), async (req, res) => {
   const user = req.user
   if (!user) return respondWithError(res, 401, "Unauthorized user")
 
-  const userObj = new User(user._id)
+  const userObj = new User()
```
cubap, line 17: Just a note to future us that `User` probably deserves the same treatment as `Project` and the normal use would be `new User(_id)` and have a separate way to lookup by `_id` if we only have some properties to query for instead. The constructor would even be able to check for the RERUM prefix and strip to the hex if present so User can be instantiated with agent or id.

utilities/getHash.mjs
```diff
@@ -0,0 +1,10 @@
+export default function getHash(agent) {
+  if (agent) {
+    return agent.split("id/").pop()
+  } else {
+    return new Error({
+      status: 404,
```
cubap, line 6: HTTP codes should be reserved for HTTP Requests

cubap (approved):
Comments, but no breaking reservations. I have not tested the error paths yet, so I won't merge it until I come back after my other tasks.

classes/Project/Project.mjs
jsvoo: Yes, we have the agent/user on the request, I'm sending that agent from the route to verif against the creator/groupMembers of the project in the Project class

cubap:
@jsvoo Everything looks correct

PR #128, opened 2024-08-12
Nodemailer

jsvoo:
- added sendMail function that uses Nodemailer to send emails
- added templated format to send different emails with varying email templates
- tested with gmail SMTP, sender email would have to be configured behind the hood
image
image

Test email received from a test gmail address:
image


utilities/mailer/index.mjs
```diff
@@ -0,0 +1,60 @@
...
+      from: process.env.TPEN_SUPPORT_EMAIL,
+      to: receiver.email,
+      subject,
+      html: htmlTemplate
+    }
+
+    const info = await transporter.sendMail(mailOptions)
+    return {status: 200, message: info.response}
+  } catch (error) {
+    return {status: error.status ?? 500, message: error.toString()}
```
cubap, line 55: Just to note, this error is on the mailer action itself, we do not "see" if there is a fake email put in or anything else.

cubap (approved):
Successfully tested against internal and external addresses with slumailrelay. Still some changes before merging, but we can do it.

package.json
```diff
@@ -45,7 +45,9 @@
       "manifesto.js": "^4.2.17",
       "mariadb": "^3.2.3",
       "mongodb": "^6.4.0",
-      "morgan": "^1.10.0"
+      "morgan": "^1.10.0",
+      "nodemailer": "^6.9.14",
+      "tpen3-services": "file:"
```
thehabes, line 50: `"tpen3-services": "file:"` is this supposed to be here?

thehabes (approved):
I did not really test it as my machine, on or off VPN, is different from the servers and being on campus and so it isn't really worth the trouble as whatever errors I encounter can't be trusted as real errors.

However, I still tested it just to see what it would look like.  The error throughput is good.  I think we are ready to deploy this and test it from our servers.

Yes, the test endpoint needs to go.  We can pass it forward with the merge so we have an easy way to test it from the deploy, but don't forget about it.

package.json
jsvoo: @cubap

PR #130, opened 2024-08-21
Add TPEN user to a Project

jsvoo:
**Changes implemented include**

- include permissions and roles files locally. we can move these to DB if preferred. 
- add mapping function to match permissions to roles
- get combined permissions for roles in case a user is assigned more than one role
- include a getUserByEmail method to `User`  so that user can be retrieved, added to project and notified by email
- `addMember` method which contains the main logic for adding member to `Project`
- basic invite member route `project/:id/invite-member`.

**What is left for this PR**

- [x] check user's access to project before allowing user to add members to group
- [x] add a `validateEmail` method to check invitee's email before proceeding
- [x] modify `checkUserAccess` method in project to accommodate roles and permissions
- [x] add members who are not current TPEN users to project
- [x] encrypt inviteCode
- [ ] passing tests for classes and routes


classes/Project/Project.mjs
```diff
@@ -1,6 +1,11 @@
 import dbDriver from "../../database/driver.mjs"
+import Permissions from "../../project/groups/permissions.mjs"
+import Roles from "../../project/groups/roles.mjs"
```
cubap, line 3: I support this. I like this way of doing it.

classes/Project/Project.mjs
```diff
@@ -45,7 +50,39 @@ export default class Project {
...
 
-  checkUserAccess(userAgent) {
+  async addMember(email, rolesString) {
+    try {
+      let user = await User.getByEmail(email)
+      const roles = this.parseRoles(rolesString)
+      let updatedProject
+      let message = `You have been invited to the TPEN project ${
+        this.projectData?.name
+      } with the following role(s): ${roles.join(", ")}.`
```
cubap, line 60: Not important for this PR, but I think we don't need the roles in the original "Join this project" message. I can separately think about some scenarios and suggest copy.

classes/Project/Project.mjs
```diff
@@ -45,7 +50,39 @@ export default class Project {
...
+        // We will replace this URL with the correct url
+        const url = `https://cubap.auth0.com/u/signup?invite-code=${newUser.inviteCode}`
+        updatedProject = projectData
+        user = newUser
+        message += `<p>Click the button below to get started with your project</p> 
+        <button class = "buttonStyle" ><a href=${url} >Get Started</a> </button>
+        or copy the following link into your web browser <a href=${url}>${url}</a> </p>`
+      }
+
+      await sendMail(user, `Invitation to ${this.projectData?.name}`, message)
```
cubap, line 75: We probably don't need to wait for this.

cubap: even though `sendMail()` is labelled `async` the Transporter.send method isn't, so I don't think there is actually a Promise to generate here.

cubap: The `err(err)` is a callback as I read it, in nodemailer, so that's where we could be rejecting or throwing to reach the `catch` block below.

classes/Project/Project.mjs
```diff
@@ -56,7 +93,12 @@ export default class Project {
     if (this.projectData.creator === userAgent) {
       return {
         hasAccess: true,
-        message: "User is the creator of the project and has access."
+        permissions: {
```
cubap, line 96: Ultimately, I think this will be changed to the OWNER Role, so the `creator` can "give away" a project if desired.

classes/Project/Project.mjs
```diff
@@ -67,33 +109,109 @@ export default class Project {
       }
     }
 
-    if (
-      !this.projectData.groups.members ||
-      this.projectData.groups.members.length === 0
-    ) {
+    const member = Object.values(this.projectData.groups).find(
```
cubap, line 112: I think this will be built to be `projectData.contributors`. At any rate, a Project will only have one "Group" so the `groups` seems to actually be like `groupmembers`.

cubap: Also, if the `.contributors` is a User Map, this could easily just be `const permissions = projectData.contributors[userAgent]?.permissions` and then `!permissions` means they aren't in the group.

The `action` argument is not used yet, but it could also be attempted to match with a failsafe. Then the error could just be "Not allowed" if `permissions` has length and "Not a member" if there are none.

cubap (commented):
bookmarking this review

cubap (approved):
I like how this looks. If you understand the comments and have any changes to make, I can put this in today.

cubap (approved):

Issue #129, opened 2024-08-19
project groups


git-voo:
We need to enhance the Project class to include an addMember() method that invites user to join a project's group.

- [x] 1. `addMember(email, role)` method 
- [x] > Handle Existing Users: If the user exists, add them to the project's group with the specified roles and combine permissions from all roles. send email notification afterwards
- [x] > Handle New Users (Invitations): If the user does not exist, generate an invitation token that includes the roles and send an invitation email.

**Additional Information**
**Roles and Permissions**: Ensure roles and permissions are defined in the roles.mjs and permissions.mjs files. path = project/groups/roles.mjs, project/groups/permissions.mjs 

**User Class Method**: Update the User class to include a method to retrieve users by email

PR #131, opened 2024-09-19
129 project groups

git-voo:
fixes issue #129 

1. refactored the permissions structure to follow ACTION_SCOPE_ENTITY
2. rewrite access checkers to follow this structure


cubap (approved):
I've made a few changes, but I think this is good now. If changes are all friendly, feel free to push it in.

As a note, there are a lot of new functions here that do not have unit tests. Even an AI generated blob to make sure that type checking and validity remains intact would be good to add.

PR #141, opened 2024-10-14
138 creating a new project is not making groups and contributors

cubap:


classes/Project/ProjectFactory.mjs
```diff
@@ -74,7 +74,8 @@ export default class ProjectFactory {
       })
       .then(async (project) => {
         const projectObj = new Project()
-        return await projectObj.create({...project, creator})
+        const group = await Group.createNewGroup(creator, { label: project.title ?? project.label ?? `Project ${new Date().toLocaleDateString()}` }).then((group) => group._id)
```
git-voo, line 77: The `Group `class has not been imported into `ProjectFactory `yet

classes/Group/Group.mjs
```diff
@@ -0,0 +1,117 @@
...
+        return Object.keys(this.members).filter(memberId => this.members[memberId].roles.includes(role))
+    }
+
+    async save() {
+        return database.save(this,process.env.TPENGROUPS)
+    }
+
+    static async createNewGroup(creator, payload) {
+        const { customRoles, label, members } = payload
+        if (!ownerId) {
```
git-voo, line 99: `ownerId `is not defined. Did you mean to extract `ownerId `from the payload?

git-voo (changes requested):

classes/Group/Group.mjs
cubap: This is missing some alignment. It should be creator. I'm fixing it

classes/Group/Group.mjs
```diff
@@ -0,0 +1,117 @@
+import dbDriver from "../../database/driver.mjs"
+const database = new dbDriver("mongo")
+
+export default class Group {
+    constructor(groupId) {
+        this._id = groupId
+    }
+
+    async #loadFromDB() {
+        return database.getById(this.groupId, "Group")
```
git-voo, line 10: `groupId `was instantiated as `this._id`

classes/Group/Group.mjs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    async getMembers() {
+        return this.members ?? await this.#loadFromDB()
```
git-voo, line 14: `this.members` was not initialized in the constructor, I think this may break other methods assigning attributes to members if `this.members` is `null`. it looks like it broke `getByRole` but I haven't tested other methods yet

classes/Group/Group.mjs
cubap: +

PR #146, opened 2024-10-17
Project normalization

cubap:


cubap:
Still running #alltests

thehabes (approved):
Conversation from dynamic review process on Slack that will be turned into issues.

I signed up with 'bguy' as my username and also provided my email.  My TPEN3 display name is 'thehabes1110' which I did not expect.  'bguy' is my `username` on the Auth0 user so we must be getting something else.
 ```
 if (!data.profile || !data.profile.displayName) {
  data.profile = data.profile || {}
  data.profile.displayName = data.email?.split("@")[0]
    ?? data.name
    ?? data.displayName
    ?? data.fullName
    ?? `User ${new Date().toLocaleDateString()}`
}
 ```

/my/projects does function as expected.  I have two projects, but 0 are returned
```
New projects I am creating don't have @type.  /my/projects/ is still hoping for @type
```

async find(query, collection) --  caved and didn't leave it as one parameter
```
You could { query, collection } if you wanted, but they are not the same, which is why its separate. collection is the db channel and the query is the payload. There are ways in Mongo to combine it, but not with our RERUM API

I know, it's tricky.  @type seemed like a way to connect the wires so we didn't always have to do payload AND collection.  However, that meant @type and now we don't think @type is a good idea for our objects.

We need another way to determine collection off the payload, or we can explicitly pass it as a second parameter outside the payload parameter

"users" is the only valid thing with 'profile'
"groups" is the only valid thing with 'members'
"projects" is the only valid thing with 'group'
```

npm run allTests
```
l failing tests
 FAIL  classes/Group/__tests__/Group.test.mjs
 
You can probably skip a handful of these

Won't prioritize Group tests for the Project PR. Not emotionally attached to any of them remaining in.
```

PR #151, opened 2024-10-31
Manage roles

git-voo:
Added endpoints to manage group member roles.

- [x] add new role to member
- [x] remove role from member
- [x] change member role

Opinion required: @cubap @thehabes  These endpoints are added under /project, what do you think about moving them to /groups?


cubap:
The interface doesn't know any group IDs, so the endpoint should be like /project/id/updateRoles with a user+roles payload or addRole, removeRole. We could put both IDs in the pattern but that might be trickier.

classes/Group/Group.mjs
```diff
@@ -68,12 +72,18 @@ export default class Group {
                     message: "Invalid roles"
                 }
             }
-            roles = roles.split(" ")
+            roles = roles.toUpperCase().split(" ")
```
cubap, line 75: This only changes them to UPPERCASE when they are passed in a String. `["permission"]` would be unaffected.

project/index.mjs
```diff
@@ -204,8 +204,135 @@ router.route("/:id/remove-member").post(auth0Middleware(), async (req, res) => {
...
+
+    if (!accessInfo.hasAccess) return respondWithError(res, 403, accessInfo.message)
+
+    const groupId = projectObj.data.group
+    const group = new Group(groupId)
+    await group.addMemberRoles(collaboratorId, roles)
+    await group.update()
+    res.status(200).json({ message: `Roles added to member ${collaboratorId}.` })
+  } catch (error) {
+    res.status(error.status || 500).json({ message: error.message || "Error adding roles to member." })
```
thehabes, line 230: maybe use respondWithError here

project/index.mjs
```diff
@@ -204,8 +204,135 @@ router.route("/:id/remove-member").post(auth0Middleware(), async (req, res) => {
...
+    const projectObj = await new Project(projectId)
+    const accessInfo = await projectObj.checkUserAccess(user._id, ACTIONS.UPDATE, SCOPES.ALL, ENTITIES.MEMBER)
+
+    if (!accessInfo.hasAccess) return respondWithError(res, 403, accessInfo.message)
+
+    const groupId = projectObj.data.group
+    const group = new Group(groupId)
+    group.setMemberRoles(collaboratorId, roles)
+
+    res.status(200).json({ message: `Roles [${roles}] updated for member ${collaboratorId}.` })
```
thehabes, line 259: Is there a compelling reason this needs to be JSON?  Why not just a textual message as the body?  It would be better if the message said "Roles [${roles}] have been set for member ${collaboratorId}".

project/index.mjs
```diff
@@ -204,8 +204,135 @@ router.route("/:id/remove-member").post(auth0Middleware(), async (req, res) => {
...
+  } catch (error) {
+    res.status(error.status || 500).json({ message: error.message || "Error updating member roles." })
```
thehabes, line 261: maybe use respondWithError here

project/index.mjs
```diff
@@ -204,8 +204,135 @@ router.route("/:id/remove-member").post(auth0Middleware(), async (req, res) => {
...
+    const projectObj = await new Project(projectId)
+    const accessInfo = await projectObj.checkUserAccess(user._id, ACTIONS.DELETE, SCOPES.ALL, ENTITIES.MEMBER)
+
+    if (!accessInfo.hasAccess) return respondWithError(res, 403, accessInfo.message)
+
+    const groupId = projectObj.data.group
+    const group = new Group(groupId)
+    await group.removeMemberRoles(collaboratorId, roles)
+
+    res.status(204).send()
```
thehabes, line 290: Since you give a message about what roles are added or set, you should also give a message about what roles were successfully removed.

project/index.mjs
```diff
@@ -204,8 +204,135 @@ router.route("/:id/remove-member").post(auth0Middleware(), async (req, res) => {
...
+    res.status(error.status || 500).json({ message: error.message || "Error removing roles from member." })
+  }
+})
+
+
+// Switch project owner
+
+router.route("/:projectId/switch/owner").post(auth0Middleware(), async (req, res) => {
+  const { projectId } = req.params
+  const { newOwnerId } = req.body
```
thehabes, line 302: How can I provide this new owner id?  It doesn't attempt to read from it as JSON and when I do content-type: "text/html" and provide a string ID I still get the "Provide the ID of the new owner." error.  I couldn't find a way to manually test this endpoint.

project/index.mjs
```diff
@@ -204,8 +204,135 @@ router.route("/:id/remove-member").post(auth0Middleware(), async (req, res) => {
...
+      return respondWithError(res, 400, "Cannot transfer ownership to the current owner.")
+    }
+
+    const currentRoles = await group.getMemberRoles(user._id)
+    // If user only has the OWNER role, we default them to CONTRIBUTOR before transferring ownership
+    Object.keys(currentRoles).length === 1 && await group.addMemberRoles(user._id, ["CONTRIBUTOR"])
+    await group.addMemberRoles(newOwnerId, ["OWNER"])
+    await group.removeMemberRoles(user._id, ["OWNER"])
+
+    res.status(200).json({ message: `Ownership successfully transferred to member ${newOwnerId}.` })
```
thehabes, line 331: Is there a compelling reason this needs to be JSON? Why not just a textual message as the body?

project/index.mjs
```diff
@@ -204,8 +204,135 @@ router.route("/:id/remove-member").post(auth0Middleware(), async (req, res) => {
...
+    if (!accessInfo.hasAccess) return respondWithError(res, 403, accessInfo.message)
+
+    const groupId = projectObj.data.group
+    const group = new Group(groupId)
+    await group.removeMemberRoles(collaboratorId, roles)
+
+    res.status(204).send()
+
+  } catch (error) {
+    res.status(error.status || 500).json({ message: error.message || "Error removing roles from member." })
```
thehabes, line 293: maybe use respondWithError here

project/index.mjs
```diff
@@ -204,8 +204,135 @@ router.route("/:id/remove-member").post(auth0Middleware(), async (req, res) => {
...
+
+    const currentRoles = await group.getMemberRoles(user._id)
+    // If user only has the OWNER role, we default them to CONTRIBUTOR before transferring ownership
+    Object.keys(currentRoles).length === 1 && await group.addMemberRoles(user._id, ["CONTRIBUTOR"])
+    await group.addMemberRoles(newOwnerId, ["OWNER"])
+    await group.removeMemberRoles(user._id, ["OWNER"])
+
+    res.status(200).json({ message: `Ownership successfully transferred to member ${newOwnerId}.` })
+  } catch (error) {
+    res.status(error.status || 500).json({ message: error.message || "Error transferring ownership." })
```
thehabes, line 333: maybe use respondWithError here

thehabes (commented):
I was able to successfully perform addRoles, setRoles, and deleteRoles.  I was not able to get /switch/roles to function because it did not seem like there was a way for me to provide the new owner id.

project/index.mjs
git-voo: This is what the endpoint expects as a body
```
 {
   "newOwnerId":"xyz"
 } 
```

cubap (approved):

thehabes (approved):

PR #154, opened 2024-11-05
Roles on group

cubap:
I think this is the lightest touch. There are other comments I'll add upstream

### DIFF
--- a/classes/Group/Group.mjs
+++ b/classes/Group/Group.mjs
@@ -51,10 +51,10 @@ export default class Group {
             throw err
         }
         this.data.members[memberId] = { roles: [] }
-        this.updateMember(memberId, roles)
+        this.setMemberRoles(memberId, roles)
     }
 
-    async updateMember(memberId, roles) {
+    async setMemberRoles(memberId, roles) {
         if (!Object.keys(this.data.members).length) {
             await this.#loadFromDB()
         }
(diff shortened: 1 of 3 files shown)


classes/Group/Group.mjs
```diff
@@ -133,32 +148,88 @@ export default class Group {
...
+        if (!this.isValidRolesMap(roleMap))
+            throw new Error("Invalid roles. Must be a JSON Object with keys as roles and values as arrays of permissions or space-delimited strings.")
+        for (const role in roleMap) {
+            delete this.data.customRoles[role]
+        }
+        this.update()
+    }
+
     async save() {
+        this.validateGroup()
```
cubap, line 190: This will always add back an OWNER/LEADER if missing, so we don't have to check upstream and will just avoid unforced errors.

classes/Group/Group.mjs
```diff
@@ -133,32 +148,88 @@ export default class Group {
...
+        this.update()
+    }
+
+    removeCustomRoles(roleMap) {
+        if (!this.isValidRolesMap(roleMap))
+            throw new Error("Invalid roles. Must be a JSON Object with keys as roles and values as arrays of permissions or space-delimited strings.")
+        for (const role in roleMap) {
+            delete this.data.customRoles[role]
+        }
+        this.update()
```
cubap, line 186: These are async, so something could go wrong. I would like to refactor this all to not save a Group until explicitly asked, but that requires changes in other supporting files.

PR #160, opened 2024-11-20
Add/Remove users fixes

cubap:


thehabes (approved):
Code looks okay.  Tested and confirmed functionality of invite into TPEN, invite into group, remove from group.

Noting that we need to catch this up to role management.

PR #164, opened 2024-12-10
Import project fixes

git-voo:


thehabes (approved):
The import endpoint functions as expected.

I noticed the resulting Project that went into the private tpen.projects collection have `@context` and `@type`.

classes/Project/ProjectFactory.mjs
```diff
@@ -37,6 +37,7 @@ export default class ProjectFactory {
     newProject.label = manifest.label
     newProject.metadata = manifest.metadata
     newProject["@context"] = "http://t-pen.org/3/context.json"
```
thehabes, line 39: hmm

classes/Project/ProjectFactory.mjs
```diff
@@ -37,6 +37,7 @@ export default class ProjectFactory {
...
+    newProject["@type"] = "Project"
```
thehabes, line 40: hmm

PR #168, opened 2025-01-14
Packages Update + Hotfix

thehabes:


git-voo (approved):
Looks good!
I pulled this down and the app runs with the changes. Other routes work as expected because they already follow the `router.route.[METHOD]` format.

Modified routes also work as expected

PR #171, opened 2025-01-21
update metadata

git-voo:


classes/Project/Project.mjs
```diff
@@ -141,6 +141,19 @@ export default class Project {
     }
   }
 
+  async updateMetadata(newMetadata) {
+    this.data.metadata = newMetadata
+    return await this.update()
+  }
+
+  async update() {
+    return await database.update({ ...this.data, type: "Project" })
```
cubap, line 150: It is probably worth making the change to `update()` that does not require `@type: Project` https://github.com/CenterForDigitalHumanities/TPEN-services/blob/9314bd0ffee370a2260b3a7abee5f896716d1a8d/database/mongo/controller.mjs#L217 The change was already made to `save()`

classes/Project/Project.mjs
```diff
@@ -141,6 +141,19 @@ export default class Project {
...
+  }
+
+  async save() {
+    return await database.save(this.data, "Project")
```
cubap, line 154: As above, you can just pass "Project" instead of adding it to the object being saved.

project/index.mjs
```diff
@@ -433,6 +433,34 @@ router.post('/:projectId/removeCustomRoles', auth0Middleware(), async (req, res)
     respondWithError(res, error.status ?? 500, error.message ?? 'Error removing custom roles.')
   }
 })
+ 
+// Update Project Metadata
+router.route("/:projectId/update-metadata").put(auth0Middleware(), async (req, res) => {
```
cubap, line 438: Simply a `PUT :id/metadata` would imply it is being updated, I think. The route may be overnamed.

project/index.mjs
```diff
@@ -433,6 +433,34 @@ router.post('/:projectId/removeCustomRoles', auth0Middleware(), async (req, res)
...
+  const { projectId } = req.params
+  const metadata = req.body.metadata
```
cubap, line 440: We have (or will have) the ability to 'set' any unprotected property on the Project, so having to put this body inside a metadata field is extra unless there is additional data that will be needed.

To be clear, the pain point for me is that a dev thinks "I want to update the metadata". RESTfully, the dev PUTs the new value to "./update-metadata" and then also puts the content inside of a `metadata` property instead of just sending a JSON Array. A dev may feel most efficient for the path to just be "I want to update (PUT) this project's metadata (projectId/metadata) with this new content ( `[{},{},{}]` ).

cubap (changes requested):
This is great. I am willing to push it through if you reject my suggestions. It makes me want to add more to it, which is a sign we're doing it right.

classes/Project/Project.mjs
```diff
@@ -141,6 +141,19 @@ export default class Project {
     }
   }
 
+  async updateMetadata(newMetadata) {
```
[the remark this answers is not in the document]
thehabes: Yes gotta document.  What is the typeof that @param

classes/Project/Project.mjs
thehabes: Clearly recognize the two layers here.  

The class Project has a `save()` and `update()` that treat the need for `"Project"` a little differently.  At a minimum reconcile this.

The mongo controller has a `save()` and `update()` that treat the need for "Project" in a significantly different way.  Determine if it is within the scope of this PR to reconcile the difference in the mongo controller as well.

project/index.mjs
thehabes: Just try to be consistent and sane.  Compare against the other .put()s.
image

project/index.mjs
thehabes: We will always find ourselves here, it is a RESTful philosophical `undefined`.  Either way makes logical sense and there is not much guidance beyond that.  

Since we have the fixed `metadata` vocabulary we should not even entertain the possibility of another metadata-like `key`.  From a minimalistic-dodge-inherited-failure-cases point of view the body should be an Array.  Only indexes that are JSON objects will be considered and others are ignored.  The considered objects themselves must contain a `label` and `value` property and other properties are ignored.  Considered objects without both `label` and `value` are ignored.

project/index.mjs
```diff
@@ -433,6 +433,34 @@ router.post('/:projectId/removeCustomRoles', auth0Middleware(), async (req, res)
...
+  const user = req.user
+ 
+  if (!user) {
+    return respondWithError(res, 401, "Unauthenticated request")
+  }
+
+  if (!metadata || !Array.isArray(metadata)) {
```
thehabes, line 447: > The body should be an Array.  Only indexes that are JSON objects will be considered and others are ignored.  The considered objects themselves must contain a `label` and `value` property and other properties are ignored.  Considered objects without both `label` and `value` are ignored.

PR #175, opened 2025-02-10
modify db.update to receive one param {data, collection}

git-voo:
mongo controller.update() only reads one param, hence, it didn't see the collection passed as second param. For this reason, older projects that contained `@type` were updated successfully but newer projects aren't.

since we don't want to store type on projects, I'm passing one payload as an object `{data, collection}`

Followup cleanups are required for all methods currently using  db.update() of the mongo controller.

We may also have to optimize the controller methods `determineDataType` and `discernCollectionFromType`. This might also lead to their deprecation.

### DIFF
--- a/classes/Group/Group.mjs
+++ b/classes/Group/Group.mjs
@@ -234,7 +234,7 @@ export default class Group {
 
     async update() {
         await this.validateGroup()
-        return database.update({ data: this.data, collection: process.env.TPENGROUPS })
+        return database.update(this.data, process.env.TPENGROUPS)
     }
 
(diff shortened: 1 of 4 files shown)


thehabes (approved):
This is good and we will do an overall refactor for the `collection` dependency existing in /database/driver.mjs

mepripri:
The change looks right, able to see the collaborators now
Screenshot 2025-02-10 at 1 46 21 PM

Added to the DB as well.
Screenshot 2025-02-10 at 1 56 25 PM

PR #177, opened 2025-02-19
172 cleaning up some awaitasync

cubap:
some weird diffs, but this is not too complicated.

Big thing to note: Only save/create/update is async. The other work within Projects and Groups are in memory until committed. This means they do not need to be async.

### DIFF
--- a/project/index.mjs
+++ b/project/index.mjs
@@ -51,7 +51,6 @@ router
     let { createFrom } = req.query
     let user = req.user
     createFrom = createFrom?.toLowerCase()
-
     if (!createFrom)
       return res.status(400).json({
         message:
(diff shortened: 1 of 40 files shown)


project/index.mjs
git-voo, line 1: I figured testing the routes in project/index.mjs would account for all methods in Project and Group classes. Below is a summary of the manual tests.

All Group related endpoints are failing with a similar message, seems the logic is trying to split a value before it's ready. Maybe a missing `await` somewhere

`project/`
 `:id` Successful

 `create`, Successful

`import?createFrom=UR`L Successful

`:id/invite-member` (error)
image

`:id/remove-member`
image

`collaborator/:collaboratorId/addRoles`
image

`collaborator/:collaboratorId/setRoles`
image

`
:projectId/addCustomRoles
`
image

PR #178, opened 2025-02-20
Changed Collections Parameter for Save()

mepripri:


app.mjs
```diff
@@ -31,7 +31,7 @@ let app = express()
 
 //Middleware to use
 app.use(logger('dev'))
-app.use(express.json())
+app.use(express.json({ limit: '50mb' }));
```
thehabes: hmm is this necessary?

thehabes (approved):
Merge this with honors, we figured out some real deep things.

app.mjs
thehabes: We know about it now.

PR #199, opened 2025-03-20
Endpoint to save changes for the new layer

mepripri:


mepripri:
In this PR I am trying to achieve an endpoint to save the modified layers to the existing Project in the services.

project/index.mjs
```diff
@@ -484,6 +484,32 @@ router.post('/:projectId/removeCustomRoles', auth0Middleware(), async (req, res)
   }
 })
 
+// Update Project Layers
+router.route("/:projectId/layers").put(auth0Middleware(), async (req, res) => {
```
cubap, line 488: https://github.com/CenterForDigitalHumanities/TPEN-interfaces/issues/90#issuecomment-2741432306 This comment shares what I think is a good pattern to provide here.

cubap (changes requested):
This looks good and quick, but I think it can be adjusted just a little.

cubap:
I do not think we should ever allow an empty label. It is probably appropriate to auto generate one that can be changed if it is missing or empty. Something like {project.label} Layer {index}

classes/Project/Project.mjs
```diff
@@ -140,6 +141,62 @@ export default class Project {
     }
   }
 
+  async addLayer(layer) {
```
cubap, line 144: Added a commit that shows some changes possible here.

classes/Project/Project.mjs
cubap, line 1: All the logic from here should be built into the Layer.mjs file.
Keep in mind that there is a version of the Layer (maybe `Layer.data`) that is what saves out to Tiny and looks like the changes I have here. The Class itself will have all the methods for adding and saving on it. The Project class can know about the Layers by ID and grab them easily when we need to send things back.

classes/Project/Project.mjs
```diff
@@ -158,6 +158,11 @@ export default class Project {
     return await this.update()
   }
 
+  async updateLayers(layers) {
```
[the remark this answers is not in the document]
cubap: I am even more convinced of this now.

project/index.mjs
cubap, line 1: This looks great. The project path makes sense for the API.

cubap (changes requested):
I think there are some structural changes, but it is mostly there.

classes/Project/Project.mjs
```diff
@@ -140,6 +141,63 @@ export default class Project {
...
+
+    try {
+      const responses = await Promise.all(canvases.map(canvas => fetch(canvas)))
+      const data = await Promise.all(responses.map(response => response.json()))
+      const layerAnnotationCollection = {
+        "@context": "http://www.w3.org/ns/anno.jsonld",
+        "id": `${process.env.RERUMIDPREFIX}${database.reserveId()}`,
+        type: "AnnotationCollection",
+        label,
+        pages: await Promise.all(data.map(async (canvas) => {
```
thehabes, line 156: The things of type "AnnotationCollection" does not support a `pages` property.  A type "AnnotationCollection" has the `items` property for this.  

Note that a private TPEN Mongodb Layer has `pages`.

project/index.mjs
thehabes: Yep this looks good!

cubap:
When this is through lets stage another push to `main`

project/layersAfterManifestImport.json
```diff
@@ -0,0 +1,38 @@
+[
```
thehabes, line 1: Generally this is out of scope here.  These are the resulting layers after importing https://tpen-project-examples.habesoftware.app/transcription-project/manifest.json

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
+[
```
thehabes, line 1: The results of adding a new Layer to the existing project mongoDBAtlas testTpen.projects _id 67e5955bdbc0d76c277ff6fc

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
...
+   {
+      "@id":1743099227843,
+      "@type":"Layer",
+      "pages":[
+         {
+            "id":"https://tpen-project-examples.habesoftware.app/transcription-project/page-1.json",
+            "type":"AnnotationPage",
+            "canvas":"https://tpen-project-examples.habesoftware.app/transcription-project/canvas-1.json",
+            "lines":[
+
```
thehabes, line 11: This page has 2 lines but shows 0

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
...
+            "lines":[
+               
+            ]
+         },
+         {
+            "id":"https://tpen-project-examples.habesoftware.app/transcription-project/page-3.json",
+            "type":"AnnotationPage",
+            "canvas":"https://tpen-project-examples.habesoftware.app/transcription-project/canvas-1.json",
+            "lines":[
+
```
thehabes, line 19: This page has 2 lines but shows 0

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
+[
+   {
+      "@id":1743099227843,
+      "@type":"Layer",
```
thehabes, line 4: @id and @type happening in import will go away eventually, but we will still need an id here.

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
...
+      "pages":[
+         {
+            "id":"https://tpen-project-examples.habesoftware.app/transcription-project/page-1.json",
+            "type":"AnnotationPage",
+            "canvas":"https://tpen-project-examples.habesoftware.app/transcription-project/canvas-1.json",
```
thehabes, line 9: A type "AnnotationPage" has a "target" not a "canvas"

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
...
+            "lines":[
```
thehabes, line 10: A type "AnnotationPage" has a "items" not "lines"

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
...
+   {
+      "@id":1743099227843,
+      "@type":"Layer",
+      "pages":[
+         {
+            "id":"https://tpen-project-examples.habesoftware.app/transcription-project/page-2.json",
+            "type":"AnnotationPage",
+            "canvas":"https://tpen-project-examples.habesoftware.app/transcription-project/canvas-2.json",
+            "lines":[
+
```
thehabes, line 33: This page has 2 lines but shows 0

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
...
+            "lines":[
+               
+            ]
+         },
+         {
+            "id":"https://tpen-project-examples.habesoftware.app/transcription-project/page-4.json",
+            "type":"AnnotationPage",
+            "canvas":"https://tpen-project-examples.habesoftware.app/transcription-project/canvas-2.json",
+            "lines":[
+
```
thehabes, line 41: This page has 2 lines but shows 0

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
...
+                  "Page 1 Lines Transcribed"
+               ]
+            },
+            "items":[
+               
+            ],
+            "target":"https://tpen-project-examples.habesoftware.app/transcription-project/canvas-1.json",
+            "partOf":[
+               {
+                  "id":"67e595f1dbc0d76c277ff6ff",
```
thehabes, line 70: hmm what layer is "id":"67e595f1dbc0d76c277ff6ff"

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
...
+                  "Page 1 Lines Translated"
+               ]
+            },
+            "items":[
+               
+            ],
+            "target":"https://tpen-project-examples.habesoftware.app/transcription-project/canvas-1.json",
+            "partOf":[
+               {
+                  "id":"67e595f1dbc0d76c277ff6fe",
```
thehabes, line 95: Hmm what layer is "id":"67e595f1dbc0d76c277ff6ff".  It should at least be a partOf this layer.

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
...
+                  "Page 2 Lines Transcribed"
+               ]
+            },
+            "items":[
+               
+            ],
+            "target":"https://tpen-project-examples.habesoftware.app/transcription-project/canvas-2.json",
+            "partOf":[
+               {
+                  "id":"67e595f1dbc0d76c277ff700",
```
thehabes, line 120: Hmm what layer is "id":"67e595f1dbc0d76c277ff700". It should at least be a partOf this layer.

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
...
+                  "Page 2 Lines Translated"
+               ]
+            },
+            "items":[
+               
+            ],
+            "target":"https://tpen-project-examples.habesoftware.app/transcription-project/canvas-2.json",
+            "partOf":[
+               {
+                  "id":"67e595f1dbc0d76c277ff701",
```
thehabes, line 145: Hmm what layer is "id":"67e595f1dbc0d76c277ff701". It should at least be a partOf this layer.

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
...
+         {
+            "id":"https://tpen-project-examples.habesoftware.app/transcription-project/page-1.json",
+            "type":"AnnotationPage",
+            "label":{
+               "en":[
+                  "Page 1 Lines Transcribed"
+               ]
+            },
+            "items":[
+
```
thehabes, line 65: this page has 2 lines but shows 0

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
...
+         {
+            "id":"https://tpen-project-examples.habesoftware.app/transcription-project/page-3.json",
+            "type":"AnnotationPage",
+            "label":{
+               "en":[
+                  "Page 1 Lines Translated"
+               ]
+            },
+            "items":[
+
```
thehabes, line 90: this page has 2 lines but shows 0

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
...
+         {
+            "id":"https://tpen-project-examples.habesoftware.app/transcription-project/page-2.json",
+            "type":"AnnotationPage",
+            "label":{
+               "en":[
+                  "Page 2 Lines Transcribed"
+               ]
+            },
+            "items":[
+
```
thehabes, line 115: this page has 2 lines but shows 0

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
...
+         {
+            "id":"https://tpen-project-examples.habesoftware.app/transcription-project/page-4.json",
+            "type":"AnnotationPage",
+            "label":{
+               "en":[
+                  "Page 2 Lines Translated"
+               ]
+            },
+            "items":[
+
```
thehabes, line 140: this page has 2 lines but shows 0

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
...
+               {
+                  "id":"67e595f1dbc0d76c277ff700",
+                  "label":{
+                     "en":[
+                        "Page 2 Lines Transcribed"
+                     ]
+                  }
+               }
+            ],
+            "next":null,
```
thehabes, line 128: I think this should be page-3.json

project/layersAfterAddingNewLayer.json
```diff
@@ -0,0 +1,162 @@
...
+                  "id":"67e595f1dbc0d76c277ff6fe",
+                  "label":{
+                     "en":[
+                        "Page 1 Lines Translated"
+                     ]
+                  }
+               }
+            ],
+            "next":"https://tpen-project-examples.habesoftware.app/transcription-project/page-4.json",
+            "prev":null
```
thehabes, line 104: I think this should be page-3.json

project/layersAfterManifestImport.json
```diff
@@ -0,0 +1,38 @@
...
+    {
+      "@id": 1743099227843,
+      "@type": "Layer",
+      "pages": [
+        {
+          "id": "https://tpen-project-examples.habesoftware.app/transcription-project/page-1.json",
+          "type": "AnnotationPage",
+          "canvas": "https://tpen-project-examples.habesoftware.app/transcription-project/canvas-1.json",
```
thehabes, line 9: A type "AnnotationPage" has a "target" not a "canvas"

project/layersAfterManifestImport.json
```diff
@@ -0,0 +1,38 @@
...
+          "lines": []
```
thehabes, line 10: A type "AnnotationPage" has a "items" not "lines"

classes/Layer/__tests__/exists.test.mjs
```diff
@@ -8,25 +8,9 @@ describe('Layer Class looks how we expect it to. #Layer_exists_unit', () => {
...
-  it('configures a correct Annotation', () => {
-    expect(layer.id).toBeDefined()
-    const json = layer.asJSON()
-    expect(json).toBeInstanceOf(Object)
-    expect(json.type).toBe('Range')
-    expect(json['@context']).toBe("https://iiif.io/api/presentation/3.0/")
-    expect(json.items).toBeInstanceOf(Array)
-    expect(json.target).toBeDefined()
+    expect(layer.addLayer).toBeInstanceOf(Function)
+    expect(layer.updateLayer).toBeInstanceOf(Function)
```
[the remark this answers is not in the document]
mepripri: Yes I will comment them too.

classes/Layer/Layer.mjs
```diff
@@ -14,53 +17,112 @@ export class Layer {
...
-            type: 'Range',
-            '@context': 'https://iiif.io/api/presentation/3.0/',
-            items: this.layer.body ?? [],
-            target: this.layer.target ?? '',
-        }
-    }
+    async addLayer(layer) {
+        await this.#load()
+        const label = layer?.label ?? `${this.data.label ?? "Default"} - Layer ${Date.now()}`
+        const canvases = layer.canvases
```
cubap, line 23: @thehabes commented on this. The `layer.pages` is correct, not `.canvases`.

classes/Layer/Layer.mjs
```diff
@@ -14,53 +17,112 @@ export class Layer {
...
-    create() {
-        return Promise.resolve(new Layer())
-    }
-
-    delete() {
-        return Promise.resolve()
-    }
+        try {
+            const responses = await Promise.all(canvases.map(canvas => fetch(canvas)))
+            const data = await Promise.all(responses.map(response => response.json()))
```
cubap, line 27: We don't need to fetch all canvases, if I am following your process here.

classes/Layer/__tests__/exists.test.mjs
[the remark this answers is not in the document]
mepripri: These are needed.

classes/Layer/Layer.mjs
```diff
@@ -14,53 +17,112 @@ export class Layer {
...
-    }
-
-    fetch() {
-        return Promise.resolve(new Layer())
-    }
-
-    getPages() {
-        return Promise.resolve('Array of Page objects')
+            await Promise.all(data.map(async (canvas) => {
+              const annotationsItems = await Promise.all(canvas.annotations.map(async (annotation) => {
```
cubap, line 36: We don't need a map here, if there is a Layer provided that has `pages` on it, we can just pass the `id` and `label`. The `target` is needed, but can be extracted from just the very first annotation (with any fragment selector stripped).

cubap: To be clear, this is a weird thing to have happen period, since new Layers should indicate something new happening. It may be useful for project importing.

classes/Layer/Layer.mjs
```diff
@@ -14,53 +17,112 @@ export class Layer {
...
+              const response = await fetch(annotation.id)
+              const annotationData = await response.json()
+              const annotationItems = {
+                id: `temp${database.reserveId()}`,
+                label: Object.values(annotationData.label)[0][0],
```
cubap, line 41: This should have a better default, since annotations rarely have labels.

classes/Layer/Layer.mjs
cubap: maybe I'm not following... if you are passing in canvases, the idea is that you are making a new layer by passing in 

```
{ 
    label: "words",
    canvases: [ <URI>, ... ]
}
```

I think where we would be getting these, the Canvas object would be easier/better to pass in.

classes/Layer/Layer.mjs
```diff
@@ -14,53 +17,112 @@ export class Layer {
...
+                items: [],
```
cubap, line 42: We don't need `items` on the Layer in the db.

classes/Layer/Layer.mjs
```diff
@@ -1,9 +1,12 @@
...
-        }
+import dbDriver from "../../database/driver.mjs"
+
+const database = new dbDriver("mongo")
+
+export default class Layer {
+    constructor(_id) {
+        this._id = _id
+        this.data = null
+        this.layer = null
```
cubap, line 9: Layer.layer is a little weird, but I get it.

classes/Layer/Layer.mjs
```diff
@@ -14,53 +17,112 @@ export class Layer {
...
+                id: `temp${database.reserveId()}`,
+                label: Object.values(annotationData.label)[0][0],
+                items: [],
+                target: annotationData.target,
+              }
+              return annotationItems
+              }))
+              layerAnnotationCollection.pages.push(...annotationsItems)
+            }))        
+            this.layer.push(layerAnnotationCollection)
```
cubap, line 49: I don't think this is already an Array?

classes/Layer/Layer.mjs
```diff
@@ -14,53 +17,112 @@ export class Layer {
...
-        return Promise.resolve(new Layer())
-    }
-
-    getPages() {
-        return Promise.resolve('Array of Page objects')
+            await Promise.all(data.map(async (canvas) => {
+              const annotationsItems = await Promise.all(canvas.annotations.map(async (annotation) => {
+              const response = await fetch(annotation.id)
+              const annotationData = await response.json()
+              const annotationItems = {
```
cubap, line 39: This variable name is hard to grok. I think it is a Page.

classes/Layer/Layer.mjs
```diff
@@ -14,53 +17,112 @@ export class Layer {
...
+    //             layer.label = label.label
+    //         }
+    //         return layer
+    //     })
+    //     this.data.layers = this.layer
+    //     return await this.update()
+    // }
+    
+    async update() {
+        return await database.update(this.data, process.env.TPENPROJECTS)
```
cubap, line 119: This writes a Layer data object to the Projects collection. This can't be right. I know it sort of works because the `this.data` fetches itself from a ProjectID, so the whole Project is in there. This should be explicitly stated somewhere or obvious from the code.

This may be a good case for using a "set" driver so we can just modify a single property on the Project.

cubap: Another easily understood pattern could be to import Layer to Project instead of the router and then let Project.addLayer(layer) alias Layer.addLayer(layer) and then send its own update.

classes/Layer/Layer.mjs
```diff
@@ -14,53 +17,112 @@ export class Layer {
...
 
-    create() {
-        return Promise.resolve(new Layer())
-    }
-
-    delete() {
-        return Promise.resolve()
-    }
+        try {
+            const responses = await Promise.all(canvases.map(canvas => fetch(canvas)))
```
thehabes, line 26: Be careful here.  It is very likely the canvas will not resolve.

classes/Layer/Layer.mjs
thehabes, line 27: Be careful here.  It is very likely a canvas from the wild will not resolve.

project/index.mjs
```diff
@@ -484,6 +485,109 @@ router.post('/:projectId/removeCustomRoles', auth0Middleware(), async (req, res)
...
+  }
+
+  try {
+    const project = new Project(projectId)
+
+    if (!(await project.checkUserAccess(user._id, ACTIONS.CREATE, SCOPES.ALL, ENTITIES.LAYER))) {
+      return respondWithError(res, 403, "You do not have permission to add layers to this project.")
+    }
+
+    const response = await new Layer(projectId).addLayer(layer)
```
cubap, line 508: This misleadingly saves a new version of the Project, comments above.

project/index.mjs
```diff
@@ -484,6 +485,109 @@ router.post('/:projectId/removeCustomRoles', auth0Middleware(), async (req, res)
...
+  }
+
+  if (!layer) {
+    return respondWithError(res, 400, "Invalid layer provided. Expected a layer object.")
+  }
+
+  try {
+    const project = new Project(projectId)
+
+    if (!(await project.checkUserAccess(user._id, ACTIONS.CREATE, SCOPES.ALL, ENTITIES.LAYER))) {
```
cubap, line 504: This loads the Project and then line 508 loads the same Project again.

project/index.mjs
```diff
@@ -484,6 +485,109 @@ router.post('/:projectId/removeCustomRoles', auth0Middleware(), async (req, res)
...
+  const { projectId, layerId } = req.params
+  const user = req.user
+  if (!user) {
+    return respondWithError(res, 401, "Unauthenticated request")
+  }
+
+  try {
+    const project = new Project(projectId)
+
+    if (!(await project.checkUserAccess(user._id, ACTIONS.DELETE, SCOPES.ALL, ENTITIES.LAYER))) {
```
cubap, line 526: Loads twice, as above

classes/Layer/Layer.mjs
thehabes, line 36: `canvas.annotations.map(async (annotationPage)` just to be clear

cubap (changes requested):
We're so close.

project/layersAfterAddingNewLayer.json
thehabes, line 1: You can delete this file now

classes/Layer/Layer.mjs
thehabes: Yes I don't think we need the check for existing AnnotationPages on the Canvases.  You have a list of canvasURIs and you know you want a temp AnnotationPage for each one (the canvasURI is their target).

classes/Layer/Layer.mjs
thehabes: This is a main sticking point.  I am wondering if we need it at all

classes/Layer/Layer.mjs
```diff
@@ -14,53 +17,112 @@ export class Layer {
...
     }
-
-    getImageLinks() {
-        return Promise.resolve('Array of image links in all pages')
+    
+    async deleteLayer(layerId) {
+        await this.#load()
+        this.layer = this.layer.filter(layer => (layer.id ?? layer["@id"]) !== `${process.env.RERUMIDPREFIX}${layerId}`)
+        this.data.layers = this.layer
+        return await this.update()
```
thehabes, line 66: the this.layer and this.data.layers have me twisted up

project/layersAfterManifestImport.json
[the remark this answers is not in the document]
mepripri: Deleted

classes/Layer/Layer.mjs
mepripri: Deleted

project/layersAfterAddingNewLayer.json
mepripri: Deleted

classes/Layer/Layer.mjs
mepripri: Removed

classes/Layer/Layer.mjs
mepripri: Removed

classes/Layer/Layer.mjs
mepripri: Added a default to page labels

classes/Layer/Layer.mjs
mepripri: Removed

classes/Layer/Layer.mjs
mepripri: Changed to Layer.data

classes/Layer/Layer.mjs
mepripri: Now it has data so yes it is.

classes/Layer/Layer.mjs
mepripri: Removed

classes/Layer/Layer.mjs
mepripri: Changed updates to Update the layers using $set.

project/index.mjs
mepripri: No more new versions created for a project using its own Project load function to fetch the layer to give it as data to the Layer Class

project/index.mjs
mepripri: Removed the redundancy

project/index.mjs
mepripri: Removed the redundancy

classes/Layer/Layer.mjs
mepripri: Removed

classes/Layer/Layer.mjs
mepripri: Cleared all. Now only this.data which has layer information

classes/Layer/Layer.mjs
mepripri: Removed

project/index.mjs
```diff
@@ -484,6 +485,113 @@ router.post('/:projectId/removeCustomRoles', auth0Middleware(), async (req, res)
...
+    const project = new Project(projectId)
+    const layers = (await project.loadProject()).layers
+
+    if (!(await project.checkUserAccess(user._id, ACTIONS.CREATE, SCOPES.ALL, ENTITIES.LAYER))) {
+      return respondWithError(res, 403, "You do not have permission to add layers to this project.")
+    }
+
+    const layer = new Layer(layers)
+    const response = await layer.addLayer(projectId, labelAndCanvases, project.getLabel())
+    res.status(200).json(response)
```
thehabes, line 511: Creating something new should have a `201` response code.

project/index.mjs
```diff
@@ -484,6 +485,113 @@ router.post('/:projectId/removeCustomRoles', auth0Middleware(), async (req, res)
...
+  try {
+    const project = new Project(projectId)
+    const layers = (await project.loadProject()).layers
+
+    if (!(await project.checkUserAccess(user._id, ACTIONS.CREATE, SCOPES.ALL, ENTITIES.LAYER))) {
+      return respondWithError(res, 403, "You do not have permission to add layers to this project.")
+    }
+
+    const layer = new Layer(layers)
+    const response = await layer.addLayer(projectId, labelAndCanvases, project.getLabel())
```
thehabes, line 510: This results in a strange response to the client. 

```
{
    "acknowledged": true,
    "modifiedCount": 1,
    "upsertedId": null,
    "upsertedCount": 0,
    "matchedCount": 1
}
```
The Layer that was created would be a more appropriate response body.


When I use the endpoint with a projectid that is bad I get an unhandled 500.  This may not need to change but it could be a little more RESTful / descriptive.
Ex.
POST http://localhost:3012/project/nope/layer/nope

project/index.mjs
```diff
@@ -484,6 +485,113 @@ router.post('/:projectId/removeCustomRoles', auth0Middleware(), async (req, res)
...
+    const project = new Project(projectId)
+    const layers = (await project.loadProject()).layers
+    
+    if (!(await project.checkUserAccess(user._id, ACTIONS.DELETE, SCOPES.ALL, ENTITIES.LAYER))) {
+      return respondWithError(res, 403, "You do not have permission to delete layers from this project.")
+    }
+
+    const layer = new Layer(layers)
+    const response = await layer.deleteLayer(projectId, layerId)
+    res.status(200).json(response)
```
thehabes, line 535: The response code for delete should be `204`

project/index.mjs
```diff
@@ -484,6 +485,113 @@ router.post('/:projectId/removeCustomRoles', auth0Middleware(), async (req, res)
...
+  try {
+    const project = new Project(projectId)
+    const layers = (await project.loadProject()).layers
+    
+    if (!(await project.checkUserAccess(user._id, ACTIONS.DELETE, SCOPES.ALL, ENTITIES.LAYER))) {
+      return respondWithError(res, 403, "You do not have permission to delete layers from this project.")
+    }
+
+    const layer = new Layer(layers)
+    const response = await layer.deleteLayer(projectId, layerId)
```
thehabes, line 534: This results in a strange response to the client. 

```
{
   "acknowledged": true,
   "modifiedCount": 1,
   "upsertedId": null,
   "upsertedCount": 0,
   "matchedCount": 1
}
```
When I use then endpoint with a layerid that is bad I get the typical success response.

Ex.
DELETE http://localhost:3012/project/67e703ca44958ad637e710d8/layer/nope

When I use the endpoint with a projectid that is bad I get an unhandled 500.  This may not need to change but it could be a little more RESTful / descriptive.
Ex.
DELETE http://localhost:3012/project/nope/layer/nope

thehabes (changes requested):
See inline comments about response code and response body.

project/index.mjs
mepripri: Handled no projectID error and response body to newLayer

project/index.mjs
mepripri: Changed to 204

project/index.mjs
mepripri: Handled no projectID error and layerID

thehabes (approved):
Tested through the manage-layers components as well as directly with TalendAPI.

These endpoints behave as expected and are guarded against weird things like invalid projects IDs and invalid layer ids.

`npm run allTests` passes.

PR #209, opened 2025-04-02
188 epic middleware to upgrade imported manifests

cubap:
not middleware, a utility. We can consider something that would be in front of some requests, but looking around, that doesn't seem to be the way to go.


mepripri:
The import structure looks awesome. I love it ❤️. I see consistency in importing two different projects. This will help me fix my layers part as now the import is consistent. It works as expected for both of them. The manage page loads with all the import details.

thehabes (approved):
I like the vault import/export and helper functions.  Successfully normalizes v2 and v3 manifests during the import process.  The result ends up in the database as expected.

Issue #176, opened 2025-02-10
Refactor db `controller` and `driver` to accomodate `collection` apropriately


git-voo:
Pending decision: How do we pass collection to db, as part of the object being updated or an extra param
Pending action: update driver actions to reflect this decision, controllers should determine how to handle collections but not drivers.
Driver actions can look like:
`action.update(data)`?
Controller actions can look like this
`action.update(data, collection)`?

thehabes:
**Current Driver Example**
```javascript
/**
 * Do an action with the provided data
 * @param data JSON data
 * @param A string designating a mongodb collection 👎 
 */
async action(data, collection) {
    ...
    return this.controller.action(data, collection).catch(err => err)
}
```

**Current Mongo Controller Example**
```javascript
  /**
   * Do something with the data document in the MongoDB
   * @param data JSON data
   * @param A string designating a mongodb collection 👎 
   * @return The inserted document JSON or error JSON
   */
  async action(data, collection) {
     ...
     return this.db.collection(collection).insertOne(data)
  }
```

**Desired Driver Example**
```javascript
/**
 * Do an action with the provided data
 * @param data JSON data
 */
async action(data) {
    ...
    return this.controller.action(data).catch(err => err)
}
```

**Current Mongo Controller Example**
```javascript
  /**
   * Do something with the data document in the MongoDB
   * @param data JSON data
   * @return The inserted document JSON or error JSON
   */
  async action(data) {
     const collection = discernCollectionFromData(data) 👀 
     ...
     return this.db.collection(collection).insertOne(data)
  }
```

cubap:
I can appreciate using the data to determine the collection. If this were meant to accommodate SQL, it would need to do the same for tables.

Casting this forward a bit, it also suggests that the same way there isn't a "Group" class for the Interface, we may wind up identifying some collections in the driver that we do not expect to directly interact with. That isn't in scope of this, but it is worth saying out loud at this point.

PR #211, opened 2025-04-02
139 factor out type type dependencies

cubap:
Move collection logic up into the driver from the mongo controller. If this is weird later when we aren't using Mongo, changign the names should accommodate tables etc.

### DIFF
--- a/database/driver.mjs
+++ b/database/driver.mjs
@@ -5,7 +5,7 @@
  * No matter which controller they choose, they have access to regularized unit functionality.
  * Additional controllers can be included in the /database directory as necessary.
  * 
- * @author Bryan Haberberger
+ * @author Bryan Haberberger, cubap
  * https://github.com/thehabes 
  * 
  */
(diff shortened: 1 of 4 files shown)


cubap:
fixes #176 as well, allowing for either format.

database/driver.mjs
```diff
@@ -161,3 +183,59 @@ class dbDriver {
...
+}
+
+/**
+ * Determine the collection name based on data type.
+ * 
+ * @param type A type string, such as "Project", or null
+ * @return the corresponding collection name from environment variables
+ */
+function discernCollectionFromType(type) {
+  if (!type) return null
```
cubap, line 211: I'm not sure if we want to be `null` here, but just `undefined` could mean a couple things, I guess.

thehabes (approved):
It could get a little weird if we start plugging in other database schemas, but the `collection` is a corollary to a thing like `table`.

Tested it with a TPEN-Interface staged to use it.  I had no issues creating new data or editing existing data.

PR #215, opened 2025-04-03
Save annotations

mepripri:


classes/Layer/Layer.mjs
```diff
@@ -1,84 +1,146 @@
...
+        if (!existingLayer) {
+            throw new Error(`Layer not found in RERUM: ${this.id}`)
+        }
+        updatedLayer = { ...existingLayer, ...layerAsCollection }
+        await databaseTiny.overwrite(updatedLayer)
+        return this
     }
-}
\ No newline at end of file
+}
```
mepripri, line 146: This function looks clean and is easy to understand.

classes/Layer/Layer.mjs
```diff
@@ -1,84 +1,146 @@
...
-                      
-            this.data.layers.push(newLayer)
-            await this.update()
-            return newLayer
-        } catch (error) {
-            console.error('Error fetching data:', error)
+    async delete() {
+        if(this.#tinyAction === 'update') {
+            await databaseTiny.remove(this.id)
+            .catch(err => false)
```
mepripri, line 71: Do we actually delete the Layer in the Project? We have a delete button in the interface. Same with the pages, is it like deleted as updates or there is a specific delete?

cubap: In Layer class, delete only takes it out of the RERUM db if it is there. Above the Layer class (in Project) it is just discarded (once this delete is effectuated).

cubap (approved):

PR #221, opened 2025-04-08
220 services for bug reporting and feedback

cubap:
I'll have to test this at work, where I've got all my tokens and such


cubap:
Hard to test, but I used a local gmail mailer to get these in: https://github.com/CenterForDigitalHumanities/TPEN-Static-Dev/issues

cubap:
We'll need to decide what repo we want these to drop into, or expect that the Interface will have to explain where it is going.

cubap:
When this fails, it usually does so quietly. If stuff is sent incorrectly, it just returns a `204` since there was no content and I don't need bots hammering it thinking there is a secret hidden inside. It also requires auth, since we should know the person.

feedback/feedbackController.js
```diff
@@ -0,0 +1,34 @@
+import { createGitHubIssue } from './githubService.js'
+import { respondWithError } from "../utilities/shared.mjs"
+
+export async function submitFeedback(req, res) {
+  const user = req.user ? `${req.user.profile.displayName} (${req.user._id})` : 'Anonymous'
+  const { page, feedback } = req.body
+
+  if (!feedback) {
+    return res.status(204).send()
```
thehabes, line 9: Ah cause this send() won't stop the rest of the code from executing.  Probably no real difference in returning this vs using `return` on the next line.

feedback/githubService.js
```diff
@@ -0,0 +1,23 @@
...
+    body: JSON.stringify({
+      title,
+      body,
+      labels: [label]
+    })
+  })
+
+  if (!response.ok) {
+    const error = await response.json()
+    throw new Error(`GitHub API error: ${error.message}`)
```
thehabes, line 21: Initially I did not trust this.  However, I messed with it a few different ways and GitHub did make errors that made sense and those bubbled to the client nicely.

feedback/feedbackController.js
```diff
@@ -0,0 +1,34 @@
+import { createGitHubIssue } from './githubService.js'
+import { respondWithError } from "../utilities/shared.mjs"
+
+export async function submitFeedback(req, res) {
```
thehabes, line 4: Maybe just a little documentation here

feedback/feedbackController.js
```diff
@@ -0,0 +1,34 @@
...
+
+  try {
+    await createGitHubIssue('Feedback', `Feedback from ${user}`, `Page: ${page}\n\nFeedback: ${feedback}`)
+    res.status(200).json({ message: 'Feedback submitted successfully' })
+  } catch (error) {
+    respondWithError(res, error.status ?? 500, error.message ?? 'Failed to submit feedback')
+  }
+}
+
+export async function submitBug(req, res) {
```
thehabes, line 20: Maybe just a little documentation here.

feedback/githubService.js
```diff
@@ -0,0 +1,23 @@
+export async function createGitHubIssue(label, title, body) {
```
thehabes, line 1: Some documentation here for sure.  A link to the GH docs where you figured out how to do this would be nice.

feedback/index.js
```diff
@@ -0,0 +1,40 @@
...
+
+  try {
+    await createGitHubIssue('Bug Report', `Bug reported by ${user}`, `Page: ${page}\n\nBug: ${bugDescription}`)
+    res.status(200).json({ message: 'Bug report submitted successfully' })
+  } catch (error) {
+    res.status(500).json({ error: 'Failed to submit bug report' })
+  }
+})
+
+module.exports = feedbackRouter
```
thehabes, line 40: This is the only place in the code where `module.exports` is used...can it just be `export`ed like we do everywhere else?

utilities/mailer/index.mjs
```diff
@@ -36,12 +36,37 @@ export const sendMail = async (email, subject, message) => {
   htmlTemplate = htmlTemplate.replace("{{subject}}", subject)
   htmlTemplate = htmlTemplate.replace("{{messageBody}}", message)
 
+  // Stop execution if any of the required environment variables are not set
```
thehabes, line 39: I like this, good idea.

Just noting it does not impact /beta/bug or /beta/feedback, but it probably isn't meant to.

thehabes (approved):

feedback/feedbackController.js
cubap: I use return on the next line for next() calls, but for sending the response, it feels more normal this way. Happy to move it.

feedback/githubService.js
cubap: nah, not for something we are going to throw away. This isn't really intended for contribution or extension

feedback/githubService.js
cubap: I was rewriting them all originally, but I found the same. Also, since we only use it within this module, we will always send it what we know to be properly formatted.

PR #223, opened 2025-04-14
API call to Update Profile

mepripri:


userProfile/privateProfile.mjs
```diff
@@ -20,6 +20,18 @@ router.route("/profile").get(auth0Middleware(), async (req, res) => {
   }
 }).all((req, res)=>    respondWithError(res, 405, "Improper request method. Use GET instead"))
 
+router.route("/profile/:id/update").put(auth0Middleware(), async (req, res) => {
```
cubap, line 23: we don't need `:id/` here because we know who it is.

userProfile/privateProfile.mjs
```diff
@@ -20,6 +20,18 @@ router.route("/profile").get(auth0Middleware(), async (req, res) => {
...
+  const user = req.user
+  if (!user) return respondWithError(res, 401, "Unauthorized user")
+  try {
+    const userObj = new User(user._id)
+    const userProfile = await userObj.updateProfile(req.body)
+    res.status(200).json(userProfile)
+  } catch (error) {
+    respondWithError(res, error?.status??500, error?.message??error.toString())
```
cubap, line 31: I think `??` needs spaces around it.

userProfile/privateProfile.mjs
cubap: we can also discard the `update` because this can just be the PUT method (tuck it in above before the `.all`)

classes/User/User.mjs
```diff
@@ -18,6 +18,26 @@ export default class User {
     return this
   }
 
+  async updateProfile(data) {
+    if(!data) throw new Error("No data provided")
```
cubap, line 22: We already have `getSelf()` so we don't need this.

classes/User/User.mjs
```diff
@@ -18,6 +18,26 @@ export default class User {
...
+
+    const existingUser = await database.findOne({ email: data.email }, "users")
+    if (existingUser && existingUser.email === data.email) {
```
cubap, line 25: I understand the concern over collision. If we are validating the input, we should find a good conflict HTTP Error code. Also, the user should have a unique `_sub`. 

*However* this is just the profile, so none of that matters. Only the `profile` property should be touched.

classes/User/User.mjs
```diff
@@ -18,6 +18,26 @@ export default class User {
...
+      throw new Error(`User with email ${data.email} already exists`)
+    }
+
+    const existingDisplayName = await database.findOne({ profile: { displayName: data.name } }, "users")
+    if (existingDisplayName && existingDisplayName.profile.displayName === data.name) {
```
cubap, line 30: I'm fine with overlapping displayNames. They don't have to be unique.

classes/User/User.mjs
```diff
@@ -18,6 +18,26 @@ export default class User {
...
+      throw new Error(`User with name ${data.name} already exists`)
+    }
+    
+    this.data = await (this.data ?? this.#loadFromDB().then(u => u.data))
```
cubap, line 34: `getSelf()` ☝️

classes/User/User.mjs
```diff
@@ -18,6 +18,26 @@ export default class User {
...
+    this.data.email = data.email
+    this.data.profile.displayName= data.name
```
cubap, line 36: ooo. I don't like this part.
Whatever is sent in as the body should be the content of the `profile` property (sanitized). We can publish an API that says we plan to surface publicly special things like "links" or "displayName" or "identities" or "homepage"

cubap (changes requested):
This is looking good. I think the changes I suggested should be helpful and easy to implement.

cubap (approved):

PR #225, opened 2025-04-23
Dashboard service for newest, last modified, last opened projects

git-voo:
This PR centralizes all of the “newest / last modified / last opened” project‐lookup logic into a new `ProjectStats` class and creates the `/projects/dashboard` route as a single service call. It also records “last opened” timestamps in ProjectStats.recordOpen whenever a user fetches a project using `project/:id`.
Returned objects contain projectId and label.

### DIFF
--- a/classes/Project/ProjectStats.mjs
+++ b/classes/Project/ProjectStats.mjs
@@ -1,0 +1,41 @@
+// classes/Project/ProjectStats.mjs
+import dbDriver from "../../database/driver.mjs"
+
+const database = new dbDriver("mongo")
+const STATS_COLL =  process.env.TPEN_USER_PROJECT_STATS
+
+export default class ProjectStats {
+    /**
+     * Record that `userId` just opened `projectId`.
+     * If a stat exists, update its lastOpened; otherwise insert a new one.
+     */
+    static async recordOpen(userId, projectId) {
+        const query = { userId, projectId }
+        const now = new Date()
+
+        // see if there's already a stat doc
+        const existing = await database.findOne(query, STATS_COLL)
+
+        if (existing) {
+            existing.lastOpened = now
+            // update by _id under the hood
+            await database.update(existing, STATS_COLL)
+            return existing
+        } else {
+            const stat = { userId, projectId, lastOpened: now }
+            await database.save(stat, STATS_COLL)
+            return stat
+        }
+    }
+
+    /**
+     * Fetch up to `limit` stats for this user, sorted by lastOpened descending.
+     * We pull all then sort/limit in memory.
+     */
+    static async getLastOpened(userId, limit = 1) {
+        const stats = await database.find({ userId }, STATS_COLL)
+        // sort descending by lastOpened timestamp
+        stats.sort((a, b) => new Date(b.lastOpened) - new Date(a.lastOpened))
+        return stats.slice(0, limit)
+    }
+}
--- a/classes/Project/ProjectStats.mjs
+++ b/classes/Project/ProjectStats.mjs
@@ -1,41 +1,76 @@
-// classes/Project/ProjectStats.mjs
 import dbDriver from "../../database/driver.mjs"
+import User from "../User/User.mjs"
 
 const database = new dbDriver("mongo")
-const STATS_COLL =  process.env.TPEN_USER_PROJECT_STATS
+const PROJECT_COLL = process.env.TPENPROJECTS
+const STATS_COLL = process.env.TPEN_USER_PROJECT_STATS
 
 export default class ProjectStats {
-    /**
-     * Record that `userId` just opened `projectId`.
-     * If a stat exists, update its lastOpened; otherwise insert a new one.
-     */
+
+    // upsert lastOpened 
     static async recordOpen(userId, projectId) {
-        const query = { userId, projectId }
         const now = new Date()
-
-        // see if there's already a stat doc
-        const existing = await database.findOne(query, STATS_COLL)
-
+        const existing = await database.findOne({ userId }, STATS_COLL)
         if (existing) {
+            existing.projectId = projectId
             existing.lastOpened = now
-            // update by _id under the hood
-            await database.update(existing, STATS_COLL)
-            return existing
-        } else {
-            const stat = { userId, projectId, lastOpened: now }
-            await database.save(stat, STATS_COLL)
-            return stat
+            return database.update(existing, STATS_COLL)
         }
+        return database.save({ userId, projectId, lastOpened: now }, STATS_COLL)
     }
 
-    /**
-     * Fetch up to `limit` stats for this user, sorted by lastOpened descending.
-     * We pull all then sort/limit in memory.
-     */
-    static async getLastOpened(userId, limit = 1) {
-        const stats = await database.find({ userId }, STATS_COLL)
-        // sort descending by lastOpened timestamp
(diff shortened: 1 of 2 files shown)


cubap:
I haven't found the ticket for this issue, so feel free to move this documentation to it when it is located.

cubap:
We are looking for a few loggable or easily gettable references:

1. Newest Project
2. Most Recently Modified
3. My Most Recent Change

## Principles

Let's set some definitions. By "new" we are looking at when something was created. In the case of a Project, it is created when it is imported or generated by the Owner. By "recent" we mean when something was modified last. In looking at modifications, a substantial change to a Project is not when users are added, metadata is updated, etc. but when Annotations are added/modified or when Pages are reordered, this is significant.

## Detectors

Given that we already have a way to get a User's Projects, this is the set from which all the other selections will be made.

For **Newest** Project, it is uncomplicated to check the `_createdAt` property on the relatively short list of Projects in hand. This property is set when the Project is created and will never change. As a timestamp, it is trivial to find the most recent.

For **Most Recently Modified** the target is not explicitly a Project, but we can consider when Page.update() would be called. The TinyPen driver is called to update (overwrite) the Annotation Page when changes are made. These changes may be for several reasons:

* A temporary Page is being upgraded and saved to RERUM because
    * an Annotation is added to its `items` that has a RERUM address
    * an existing Annotation is updated and gets a new RERUM id, which must be swapped in `Page.items`
* An existing Page with a RERUM id is being overwritten because
    * Annotations in `items` change, as above
    * Page order is shifted, changing the `prev` and `next` properties 

When these changes are made in RERUM, the `__rerum.modifiedAt` and `__rerum.isOverwritten` both receive a timestamp. It is very expensive, however, to search through all possible elements in RERUM proper. Although the most recent change may be considered to be a single Line, the Page is probably a reasonable scope.

For **My Most Recent Change** all the similar indicators from above apply, but with the added complication of checking that the user matches the creator or modifier. There is no good way to check all these cases, or even to understand how a situation like this would play out:

> User A creates Lines 1-8, in order, on three consecutive Pages 1-3. All these Pages were recently changed by A. User A adds two more Lines to the end of Page 2. All these Pages were recently changed by A. Following that , User B changes the first Line of Page 2 and 3. Now each Page was recently changed by B.
>
> There is no record of User A changing any Page except 1 and we'll need to either return an earlier change or drill down to the contained Annotations to find a contribution from the User.

Should the most recent change be:

* Page 2, which was the last change the User actively made, even though there were changes made after?
* Page 1, which still shows the User as the last modification?
* Page 3, which was the Page most recently created by the User?

cubap:
## Suggestions

These considerations should be implemented to support the behaviors needed here.

### Attach to Projects

When `/my/projects` is returned now, it contains a simple array of Projects. This may be the most reasonable response to hijack. Instead of 

```json
[ {
    "_id": "1234hex",
    "label": "First Project",
    "roles": ["LEADER","CONTRIBUTOR"]
  }, {
    "_id": "4321hex",
    "label": "Second Project",
    "roles": ["CONTRIBUTOR"]
  }, {
    "_id": "9949hex",
    "label": "Third Project",
    "roles": ["VIEWER"]
}]
```

We can consider one these alternatives:

```json
[ {
    "_id": "1234hex",
    "label": "First Project",
    "roles": ["LEADER","CONTRIBUTOR"],
    "_createdAt": "timestamp",
    "_modifiedAt": "timestamp",
    "_updates": {
        "latest": "https://store.rerum.io/v1/id/222",
        "userA-id": "https://store.rerum.io/v1/id/111",
        "userB-id": "https://store.rerum.io/v1/id/222",
        "userC-id": "https://store.rerum.io/v1/id/111"
    }
  }, {
    "_id": "4321hex",
    "label": "Second Project",
    "roles": ["CONTRIBUTOR"],
    "_createdAt": "timestamp",
    "_modifiedAt": "timestamp",
    "_updates": {
        "latest": "https://store.rerum.io/v1/id/777",
        "userA-id": "https://store.rerum.io/v1/id/777",
        "userB-id": "https://store.rerum.io/v1/id/444",
        "userC-id": "https://store.rerum.io/v1/id/777"
    }
  }, {
    "_id": "9949hex",
    "label": "Third Project",
    "roles": ["VIEWER"],
    "_createdAt": "timestamp",
    "_modifiedAt": "timestamp",
    "_updates": {
        "latest": "https://store.rerum.io/v1/id/888",
        "userA-id": "https://store.rerum.io/v1/id/888",
        "userB-id": "https://store.rerum.io/v1/id/888",
        "userC-id": "https://store.rerum.io/v1/id/888"
    }
}]
```

or, as a separate object that doesn't need processing on the client:

```json
{ "metrics": {
    "newest": "https://store.rerum.io/v1/id/888",
    "mostRecent": "https://store.rerum.io/v1/id/777",
    "myRecent": "https://store.rerum.io/v1/id/111"
  },
  "projects": [ {
      "_id": "1234hex",
      "label": "First Project",
      "roles": ["LEADER","CONTRIBUTOR"],
      "_createdAt": "timestamp",
      "_modifiedAt": "timestamp",
      "_updates": {
          "latest": "https://store.rerum.io/v1/id/222",
          "userA-id": "https://store.rerum.io/v1/id/111",
          "userB-id": "https://store.rerum.io/v1/id/222",
          "userC-id": "https://store.rerum.io/v1/id/111"
      }
    }, {
      "_id": "4321hex",
      "label": "Second Project",
      "roles": ["CONTRIBUTOR"],
      "_createdAt": "timestamp",
      "_modifiedAt": "timestamp",
      "_updates": {
          "latest": "https://store.rerum.io/v1/id/777",
          "userA-id": "https://store.rerum.io/v1/id/777",
          "userB-id": "https://store.rerum.io/v1/id/444",
          "userC-id": "https://store.rerum.io/v1/id/777"
      }
    }, {
      "_id": "9949hex",
      "label": "Third Project",
      "roles": ["VIEWER"],
      "_createdAt": "timestamp",
      "_modifiedAt": "timestamp",
      "_updates": {
          "latest": "https://store.rerum.io/v1/id/888",
          "userA-id": "https://store.rerum.io/v1/id/888",
          "userB-id": "https://store.rerum.io/v1/id/888",
          "userC-id": "https://store.rerum.io/v1/id/888"
      }
  }]
}
```

cubap:
It is also possible to consider an endpoint to just get the `metrics` block from above (say `/my/metrics`) but this would still have to do all the work of getting the projects to do it.

## What to Do

When we create a Project, we set the `_createdAt`. That's the only thing needed to find the newest one. Once we get the pipeline for the User's Projects, we can either retain that property or use it to add a `newest`.

Whenever a Page is updated, a call is made through TinyPen to update it. This is an opportunity to handle the other two cases. A Page is part of a Layer/Project and will have a RERUM ID. When we update it, we should update the Project with `_lastModified` on the Project in local Mongo. This also updates `_modifiedAt` as a bonus.

```json
{ "_lastModified": "https://store.rerum.io/v1/id/888" }
```

The `_lastModified` is always changed. The rest is a map of contributor users, whoever is making the change. The value is always the URI of the Annotation Page with an update. Across all Projects, the most recent is found between `_modifedAt` and `_lastModified`.

To track a specific User's recent changes, we can follow the same process. Whenever a Page is updated, and call is made through TinyPen, we should update the User with `_lastModified` on the User in local Mongo. This value can be read directly from the User as the recently modified Page.

PR #227, opened 2025-04-25
Add project metrics to /my/projects

git-voo:
newest : project with the most recent _createdAt timestamp
lastModified:  last modified by any member of project
myRecent: last modified by me


classes/Page/Page.mjs
```diff
@@ -1,50 +1,50 @@
...
-        if(this.id.startsWith(process.env.RERUMIDPREFIX)) {
+        Object.assign(this, { id, label, target, partOf })
+        if (this.id.startsWith(process.env.RERUMIDPREFIX)) {
             this.#tinyAction = 'update'
         }
         return this
     }
 
-    static build(layerId, canvas, prev, next, lines = []) {
+    static build(projectId, layerId, canvas, prev, next, lines = []) {
```
cubap, line 37: Why is projectId here? I don't think it is used.

classes/Page/Page.mjs
```diff
@@ -1,50 +1,50 @@
...
         }
         if (!canvas || !canvas.id) {
             throw new Error("Canvas with id is required to create a Page instance.")
         }
-
         const id = lines.length
             ? `${process.env.RERUMIDPREFIX}${databaseTiny.reserveId()}`
             : `${process.env.SERVERURL}layer/${layerId.split("/").pop()}/page/${databaseTiny.reserveId()}`
-        this.data = {
+        const pageDoc = {
```
cubap, line 47: This is good and I think will conflict when you merge, but the other version is similar.

classes/Page/Page.mjs
```diff
@@ -55,7 +55,9 @@ export default class Page {
             prev,
             next
         }
-        return this
+        const page = new Page(layerId, pageDoc)
+        page.data = pageDoc
```
cubap, line 59: I don't think there is a need for `.data`, but maybe you can justify it.

classes/Page/Page.mjs
```diff
@@ -90,12 +91,15 @@ export default class Page {
...
-     * @returns {Promise} Resolves to the updated Layer object as stored in Project.
-     */
-    async update() {
-        if (this.#tinyAction === 'update' || this.items.length) {
-            await this.#setRerumId().#savePageToRerum()
+      * Check the Project for any RERUM documents and either upgrade a local version or overwrite the RERUM version.
+      * @returns {Promise} Resolves to the updated Layer object as stored in Project.
+      */
+    async update(userId) {
+        const hasItems = Array.isArray(this.data?.items) && this.data.items.length > 0
```
cubap, line 98: What are these changes for? I'm not sure it wasn't working before. Page.update shouldn't save itself to the Project, just RERUM.

classes/Page/Page.mjs
```diff
@@ -115,11 +119,43 @@ export default class Page {
...
+        if (this.#tinyAction === 'update') {
             // associated Annotations in RERUM will be left intact
-            await databaseTiny.remove(this.id)
-            .catch(err => false)
+            await databaseTiny.remove(this.id).catch(err => false)
         }
         return true
     }
+
+    async #recordModification(userId) {
```
cubap, line 129: This seems to be tryingt o do a lot of work that doesn't need to be done except in Project.

cubap: Maybe this should eb somethign passed onto User?

cubap (changes requested):
This is very close. At least read my thoughts. We'll have to get the mjs changes carried over as well, which won't be a thrill.

cubap:
image

conflicts resolved. merged upstream

cubap (approved):

PR #240, opened 2025-05-01
235 save annotations

cubap:
* POST to create a new Line /project/:pid/page/:pid/line
* PUT to update a Line /project/:pid/page/:pid/line/:lid

If there is a `body` the line should get a RERUM id and that in turn should upgrade any temp id in a page or layer to a RERUM id.


cubap:
> - when I POST a Line (with body content) to a temporary Page that is part of a temporary Layer the Layer does not end up with a RERUM id.  The Page and Line do have a RERUM id and the Page is still partOf the temp Layer.  See https://devstore.rerum.io/v1/id/681e5d9dce2f2301f3c37035

I think this is correct, at least for now. The Layer logic needs to reach down into the Pages to update itself if it is significantly changed. The use of overwrite at this point on the Page prevents these significant changes. I think we can move that line in the future, but it is functional where it is.

> - I am able to POST a Line that is not type: Annotation and it saves to RERUM that way.  See https://devstore.rerum.io/v1/id/6824d6ccccafd8def2305652

Until we have strong opinions about how a Line is shaped, this is a legal Line. There may be subtypes of Annotation that we want to support so I just pass that through for now. It should return it that way as well, so the Interface can have opinions we do not.

> - I am able to PUT a Line that is not type:Annotation and it saves to RERUM that way in such a way that the new object is no longer a type: Annotation.  See https://devstore.rerum.io/v1/id/6824d9e8cfd83da30865d4d2

As above, our default object is `type:Annotation` but not enforced if it is intentionally changed.

> - A line is "@context":"[http://iiif.io/api/presentation/3/context.json"](http://iiif.io/api/presentation/3/context.json%22).  A Layer and a Page are "@context": "http://www.w3.org/ns/anno.jsonld".  We probably mean to be consistent.

Correct and consistent are very close here. The Page has no non-WebAnnotation things to claim, so it can be anno.jsonld. A Line may often use IIIF-y things like custom motivations, so it needs to have presi-3. Since WebAnnotation is in Presi-3, we can set it all to Presi-3, but it is sotr of "overcorrect". /🤷🏽‍♂️

PR #244, opened 2025-05-08
Adding Tools to Project

mepripri:
The endpoint does not require authentication.

To select which tools the user needs just change status using checkbox.
Here is the endpoint : /project/:projectID/tools?action=updatetools

Body is : { [
  { value: 'page', state: false },
  { value: 'inspector', state: true },
  { value: 'characters', state: false },
  { value: 'xml', state: false },
  { value: 'fullpage', state: true },
  { value: 'history', state: false },
  { value: 'preview', state: true },
  { value: 'parsing', state: true },
  { value: 'compare', state: true },
  { value: 'rtl', state: true },
  { value: 'cappelli', state: false },
  { value: 'enigma', state: true },
  { value: 'latin', state: true },
  { value: 'vulgate', state: false }
] }

To add a tool with iframe
Here is the endpoint : /project/:projectID/tools?action=addtools

Body is : { [ { name: '', value: '', url: '', state: true } ] }


thehabes (changes requested):
Creating and Updating tools should not be the same method.

POST should create tools.

PUT should update tools.

Then there should be no ?action= url parameter since what can be done is separate by what method is used.

It looks like the end point is looking for a body like

```JSON
{
  "tools":[
    {
        "name": "another",
        "value": "another",
        "url": "another",
        "state": true
    }
  ]
}
```

Can it be changed just to look for a body like the following?

```JSON
[
    {
        "name": "another",
        "value": "another",
        "url": "another",
        "state": true
    }
]
```

Make sure to have request body validation and appropriate responses.  For example, this body gave me the response "at least one tool is required" but I have a good one in there!

```JSON
{
  "tools":[
    {
        "nam": "another3",
        "valu": "another3",
        "url": "another3",
        "state": true
    },
    {
        "name": "another4",
        "value": "another4",
        "url": "another4",
        "state": true
    }
  ]
}
```

I am also able to add a tool with the same name and the same url as many times as I would like.  We should probably try to stop duplicates.

cubap (approved):
This looks great.

cubap (approved):

cubap (approved):

thehabes (approved):

PR #245, opened 2025-05-14
230 clean up project routers

cubap:
Take a very long projectrouter and break it into logical files that are easier to maintain.


cubap:
tests pass but not `allTests`. Some manual digging is needed to see what might be awful.

cubap:
image
 
trada!

thehabes (approved):
Code re-org looks good.  I used a recent TPEN Interfaces to go through and do all the project-y things that moved around here and there were no API errors.

Code itself is good and cleaned up.

Didn't look too hard at tests, since we know we need to make a run at them.  npm run allTests does pass locally.

PR #257, opened 2025-05-28
256 import creates bad ids

cubap:


thehabes (approved):
Tested and confirmed the ids being generated for Layer and Page from a Manifest import are correct

PR #266, opened 2025-06-20
IIIF Image Service

mepripri:


classes/Project/ProjectFactory.js
```diff
@@ -216,6 +216,71 @@ export default class ProjectFactory {
...
+        region === "full" ||
+        region === "square" ||
+        /^pct:\d+(\.\d+)?,\d+(\.\d+)?,\d+(\.\d+)?,\d+(\.\d+)?$/.test(region) || 
+        /^\d+,\d+,\d+,\d+$/.test(region)
+      )
+    }
+
+    function isValidIIIFSize(size) {
+      return (
+        size === "max" ||
```
cubap, line 235: in a 2.0 image API, "max" might be called "full". You can see why they changed it to avoid confusion.

classes/Project/ProjectFactory.js
```diff
@@ -216,6 +216,71 @@ export default class ProjectFactory {
...
+
+    function isValidIIIFQuality(quality) {
+      return (
+        quality === "default" ||
+        quality === "color" ||
+        quality === "gray" ||
+        quality === "bitonal"
+      )
+    }
+
```
cubap, line 265: I was not able to break these tests with any type 1 or 2 errors.

classes/Project/ProjectFactory.js
```diff
@@ -216,6 +216,71 @@ export default class ProjectFactory {
...
+            throw new Error(`Failed to fetch IIIF info: ${response.statusText}`)
+          }
+          return response.json()
+        })
+        .then(info => {
+          if (info?.protocol === "http://iiif.io/api/image") {
+            isIIFImage = true
+            IIIFServiceJson = info
+          }
+        })
```
cubap, line 281: You will want a `.catch(err)` at the end of this fetch chain for times when you don't get to response.ok, like a server 500.

PR #269, opened 2025-06-23
Patch text for services

cubap:
Saves incoming text patches


thehabes:
Being reviewed in #271

PR #271, opened 2025-06-24
270 multiple updates to a page in parallel discards data

cubap:


database/tiny/controller.js
```diff
@@ -184,27 +184,56 @@ class DatabaseController {
...
      */
     async overwrite(data) {
         err_out._dbaction = this.URLS.OVERWRITE
+        
+        const headers = {
+            'Content-Type': 'application/json; charset=utf-8'
+        }
+        
+        // Add optimistic locking header if __rerum.isOverwritten exists
+        if (data.__rerum?.isOverwritten) {
```
thehabes, line 199: It does seem like requests that come in here always have the __rerum property so this will essentially make every overwrite request out of here have this header.

thehabes (approved):

PR #275, opened 2025-07-14
Adding Creator

mepripri:


classes/Page/Page.js
```diff
@@ -22,18 +22,18 @@ export default class Page {
      * @param {Array} items The array of Annotation objects.
      * @seeAlso {@link Page.build}
      */
-    constructor(layerId, { id, label, target, items = [] }) {
+    constructor(layerId, { id, label, target, items = [], creator = null, prev = null, next = null }) {
         if (!id || !target) {
             throw new Error("Page data is malformed.")
         }
-        Object.assign(this, { id, label, target, partOf: layerId, items })
+        Object.assign(this, { id, label, target, items, creator, prev, next })
```
cubap, line 29: We're losing `partOf` here which I think is important.

classes/Page/Page.js
```diff
@@ -74,11 +74,12 @@ export default class Page {
...
             label: { "none": [this.label] },
-            target: this.target,
-            partOf: this.partOf,
             items: this.items ?? [],
             prev: this.prev ?? null,
-            next: this.next ?? null
+            next: this.next ?? null,
+            creator: `https://store.rerum.io/v1/id/${this.creator}`,
+            target: this.target,
+            partOf: [{ id: this.partOf, type: "AnnotationCollection" }]
```
cubap, line 86: You don't have `this.partOf` anymore

classes/Project/ProjectFactory.js
```diff
@@ -288,19 +292,22 @@ export default class ProjectFactory {
           id: `${process.env.SERVERURL}project/${copiedProject._id}/page/${database.reserveId()}`,
           label: page.label,
           target: page.target,
+          creator: copiedProject.creator,
```
cubap, line 295: The creator should be the new person cloning the Project, not the original.

classes/Project/ProjectFactory.js
```diff
@@ -288,19 +292,22 @@ export default class ProjectFactory {
...
-                target: itemData.target,
-                body: itemData.body,
-                motivation: itemData.motivation,
+                type: itemData.type,
                 label: itemData.label,
-                type: itemData.type
+                motivation: itemData.motivation,
+                body: itemData.body,
+                target: itemData.target,
+                creator: copiedProject.creator
```
cubap, line 307: The creator should be the new person cloning the Project, not the original.

classes/Project/ProjectFactory.js
```diff
@@ -288,19 +292,22 @@ export default class ProjectFactory {
...
               })
-              return await newItem.update()
+              await newItem.update()
```
cubap, line 309: If you change the `id` before updating, you will not be updating the old Line. We want to fork the Line in RERUM and then store the new `_id`. 

It may be useful to allow `update()` to take a parameter of the address to update, if you want to have the new ID already in hand. I'm not certain that we are built to do this all the way up at the moment.

classes/Project/ProjectFactory.js
```diff
@@ -288,19 +292,22 @@ export default class ProjectFactory {
...
+              return newItem.asProjectLine()
```
cubap, line 310: update already comes back as a line.

classes/Project/ProjectFactory.js
```diff
@@ -703,42 +710,65 @@ export default class ProjectFactory {
...
       label: { none: [project.label] },
       metadata: project.metadata,
-      items: await this.getManifestItems(project),
+      items: await this.getManifestItems(project, manifestJson, creator),
     }
     return manifest
   }
 
-  static async getManifestItems(project) {
+  static async getManifestItems(project, manifestJson, creator) {
```
cubap, line 732: `getManifestItems()` doesn't sound like it is about to save a _lot_ of items to RERUM.

classes/Project/ProjectFactory.js
```diff
@@ -703,42 +710,65 @@ export default class ProjectFactory {
...
+                const annotationPage = {
+                  id: page.id,
+                  type: "AnnotationPage"
+                }
+                canvasRerum.annotations.push(annotationPage)
+              }
+            })
+          }))
+          await databaseTiny.save(canvasRerum)
+          canvasItems.annotations = await this.getAnnotations(canvasRerum)
```
cubap, line 761: You already have the annotations on canvasRerum, right?

classes/Project/ProjectFactory.js
```diff
@@ -760,6 +790,8 @@ export default class ProjectFactory {
             label: annotationData.label,
             items: await this.getLines(annotationData),
             partOf: annotationData.partOf,
+            prev: annotationData?.prev?.startsWith(process.env.RERUMIDPREFIX) ? annotationData.prev : null,
```
cubap: This check isn't needed, right? It is either already `null` or it is a resolvable value. Even if we haven't saved it to RERUM (which I think it always is here), the t-pen.org link will still provide what we expect.

database/driver.js
```diff
@@ -210,6 +210,8 @@ function determineDataType(data, override) {
  if (data._sub) return "User"
  if (data.members) return "Group"
  if (data.group) return "Project"
+ if (data.total && data.first && data.last) return "Layer"
```
cubap, line 213: We do not have collections in the DB for Canvas or Layer. This shouldn't be here. In fact, Page and Line shouldn't be here either.

database/driver.js
```diff
@@ -226,6 +228,8 @@ function discernCollectionFromType(type) {
   
   switch (type) {
     case "Project":
+    case "Canvas":
```
cubap, line 231: I see what's happening here, but we will bork soem things if we have a data object that looks like a Layer  or Line and we just save it to the "Projects" collection.

layer/index.js
```diff
@@ -57,8 +58,8 @@ router.route('/:layerId')
             label ??= label ?? layer.label
             if(canvases?.length === 0) canvases = undefined
             const updatedLayer = canvases ?
-                Layer.build(projectId, label, canvases)
-                : new Layer(projectId, {id:layer.id, label, pages:layer.pages})
+                Layer.build(projectId, label, canvases, req.user._id)
```
cubap, line 61: `creator` isn't the `req.user._id` it is the agent from that token. It may be handy in the object, but can also be assembled without trouble as we worked hard to make the hexcodes match.

cubap: It is only when the creator goes to RERUM that it must be a URI, so we can change it at either end.

layer/index.js
```diff
@@ -89,7 +90,7 @@ router.route('/').post(auth0Middleware(), async (req, res) => {
 
         if (!project) return utils.respondWithError(res, 404, 'Project does not exist')
 
-        const newLayer = Layer.build(projectId, label, canvases)
+        const newLayer = Layer.build(projectId, label, canvases, req.user._id)
```
cubap, line 93: creator isn't the req.user._id it is the agent from that token. It may be handy in the object, but can also be assembled without trouble as we worked hard to make the hexcodes match.

line/index.js
```diff
@@ -61,7 +61,7 @@ router.post('/', auth0Middleware(), async (req, res) => {
     let newLine
     // This feels like a use case for /bulkCreate in RERUM.  Make all these lines with one call.
     for (const lineData of inputLines) {
-      newLine = Line.build(req.params.projectId, req.params.pageId, { ...lineData })
+      newLine = Line.build(req.params.projectId, req.params.pageId, { ...lineData }, user._id)
```
cubap, line 64: creator isn't the req.user._id it is the agent from that token. It may be handy in the object, but can also be assembled without trouble as we worked hard to make the hexcodes match.

page/index.js
```diff
@@ -40,8 +40,12 @@ router.route('/:pageId')
         target: pageObject.target,
         partOf: pageObject.partOf,
         items: pageObject.items ?? [],
-        prev: pageObject.prev ?? null,
-        next: pageObject.next ?? null
+        ...pageObject.prev?.startsWith(process.env.RERUMIDPREFIX) && {
```
cubap, line 43: same prev/next comment as above

project/projectReadRouter.js
```diff
@@ -21,7 +21,7 @@ router.route("/:id/manifest").get(auth0Middleware(), async (req, res) => {
     if (!await new Project(id).checkUserAccess(user._id, ACTIONS.UPDATE, SCOPES.ALL, ENTITIES.PROJECT)) {
       return respondWithError(res, 403, "You do not have permission to export this project")
     }
-    const manifest = await ProjectFactory.exportManifest(id)
+    const manifest = await ProjectFactory.exportManifest(id, user._id)
```
cubap, line 24: This is reading a Project, we should already have a creator on the Project items. We woudln't want to replace any of them with the person who requested the manifest be created.

cubap (approved):
It works where I expect it to. This represents a lot of work. Good job

PR #285, opened 2025-07-29
Update Copy Project Layer Fix

mepripri:


thehabes (approved):
Tested manually and project Layers/Pages are copied as expected.  Layers with content have a devstore id and layers without content has a temp id.  The devstore id and temp id resolve with the expected JSON (added a small fix for this to the PR).

PR #299, opened 2025-08-15
public-profile

mepripri:


cubap (changes requested):
I think this already has a place at `/user/:id`. The route `/my/:id/public-profile` doesn't make sense to me. You might want to merge the code together

PR #300, opened 2025-08-18
TPEN28 Fix

mepripri:


thehabes (approved):
Did some pair coding to straighten things out.  Manually tested against https://github.com/CenterForDigitalHumanities/TPEN-interfaces/pull/278 and also with TalendAPI by hitting the endpoint.  I introduced various errors that were handled by returning good status codes and error messages.

PR #298, opened 2025-08-07
295 contentchanged for layers and pages

cubap:

PR #305, opened 2025-08-19
Set line creator if not already defined

cubap:
Adds logic to assign the line's creator property to the user's agent if it is undefined when processing items. This ensures that the creator information is consistently set for new or updated lines.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1787,9 +1787,9 @@
          "license": "MIT"
       },
       "node_modules/brace-expansion": {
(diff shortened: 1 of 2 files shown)


thehabes (approved):
This worked for me and saving lines in the /annotator gives a 200.  However, so does the `development` branch.  With this change will https://github.com/CenterForDigitalHumanities/TPEN-services/pull/298 be functional on /annotator?

There is also a change to pack-lock.json in here, make sure that is at least on purpose

PR #328, opened 2025-09-17
Default Tools and New Project with custom Tools

mepripri:


thehabes (commented):
Here are some situations that should be addressed before merging the code.

**Validate the schema** of tools in the endpoint.  If there is a bad tool, drop the bad tool and continue (as opposed to throwing a 400)
image

This added the string "yep" to the array of tools in this project
image

cubap (approved):

Issue #330, opened 2025-09-18
Validate PUT /profile request bodies


thehabes:
Values for the keys provided in request bodies are not validated.  They also are not limited in any way, and I can essentially provide an infinite number of keys.  For example, the following request body succeeds in updating my profile with these values.

```JSON
{
    "imageURL": "https://avatars.githubusercontent.com/u/3287006?v=4",
    "displayName": "THEHABES",
    "name": "THEHABES",
    "orchidId": "0009-0008-6288-0257",
    "nsfId": "not an id",
    "institutionalId": 123123,
    "linkedin": "while(true) return true",
    "twitter": "not an url",
    "instagram": "not an url",
    "facebook": "not an url",
    "github": "not an url",
    "homepage": "not an url"
    "silly": "goose",
    "bryan": "dude",
    "and" : "so on"
}
```

PR #331, opened 2025-09-18
Protect PUT /profile route

thehabes:
This endpoint takes the request body wholesale and adds it into the user object in the database under `profile`.  Keys and values are not limited in any way.  For the keys we expect (such as those found on the Manage Profile interface) they are not validated in any way.  See #330


cubap (approved):
I made a change, but it fixed an edge problem with for-in

PR #381, opened 2025-11-12
Cleanup

thehabes:
Do an audit for unused imports, unused functions, unused parameters, or any other unused code that we can clean out.

### DIFF
--- a/classes/Group/Group.js
+++ b/classes/Group/Group.js
@@ -90,7 +90,7 @@ export default class Group {
             roles = roles.split(" ")
         }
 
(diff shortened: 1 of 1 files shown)


classes/Layer/Layer.js
```diff
@@ -102,12 +102,6 @@ export default class Layer {
         }
     }
 
-    #getPageReference({ id, label, target }) {
```
cubap: hm. Good catch. I think this should be used, but we'll put it back when that happens.

PR #503, opened 2026-05-01
Fix PUT /page/:pageid Column Generation Behavior

### LINKED ISSUE #502: Investigate 'one column per line' Behavior
This behavior was discovered when testing the TPEN Prompts tool.  Two prompts in particular: 'Detect Lines' as well as 'Detect Lines and Transcribe'.

## Problem

The prompts instruct the LLM to order the lines in reading order then do one PUT request to the page containing all the line data.  That single PUT request will create all the lines and update the page.

After this occurs, line-by-line navigation does not occur in order.  This is because when each line was created, a column was also generated for that line to be put in.  This caused an equal number of columns and lines.  The columns themselves are not in any particular order, but when columns are present they are used for the line-by-line navigation ordering principle.  

> If all columns are removed, line-by-line navigation will occur in order because the page `items` will be used as the ordering principle instead, which was intentionally ordered in reading order.

## Desired Behavior

It is wrong that a column is being created along with each new line.  The desired outcome is that all lines in the PUT request are created without generating any new columns.  Line creation should never automatically trigger column creation without being instructed to do so.

thehabes:
Closes #502.

PUT `/page/:pageId` was creating one new column per new line item. This was observed when the "Detect Lines" prompts submit all lines in a single PUT, which produced N single-line columns whose order had no relationship to reading order — breaking line-by-line navigation.

## Change

Removed the auto-column-create branch in the `updatedItemsList` loop and unwrapped the surrounding `else` so the split-into-existing-column logic runs unconditionally.

## Preserved behavior

- Existing-column line-id remapping when RERUM versions a changed line.
- Splitting a RERUM-tracked line: numeric date-id children are appended to the parent's column.
- Line deletion: dropped lines leave their column; empty columns are deleted.
- Explicit column creation via `POST /page/:pageId/column` is unchanged.

## Side effect

`pageColumnsUpdate.find(...)` is now `pageColumnsUpdate?.find(...)`. The new block runs for every `updatedItemsList` entry, so the variable can legitimately be `null` (column-less page). Closes a latent null-deref on column-less pages whose lines get RERUM-versioned by a PUT.

## Manual testing

| # | Scenario | Starting state | Request | Result |
|---|---|---|---|---|
| 1 | Bulk-create on a page with one column | 1 line, 1 column (`A`) | echo existing line + 2 brand-new lines | ✅ 3 items, `Column A` unchanged, no new columns |
| 2 | Single-column split | continued from 1 | parent line (body change) + 2 numeric date-id children + 2 preserved lines | ✅ 5 items; `Column A` tracked new parent URI and gained both split children |
| 3 | Line deletion from a column | continued from 2 | 4 items; one column-tracked child omitted | ✅ Column dropped the omitted child; tracked new URIs for the rest |
| 4 | No-op PUT (re-versioning guard) | reset: 1 line, 1 column | same id + same target, no body | ✅ URI unchanged — `hasAnnotationChanges` correctly skipped re-versioning |
| 5 | Multi-column split with brand-new line in same PUT | reset: 2 lines, 2 columns (`A`, `B`) | `A` (body edit) + numeric child for A; `B` (body edit) + numeric child for B; 1 brand-new line | ✅ Both columns tracked their new parent URI and gained their own child only; brand-new line created no third column |
| 6 | Bulk-create on a column-less page | different page: 1 line, 0 columns | echo existing line + 5 brand-new lines | ✅ 6 items in submission order; `columns` still `null`; existing URI preserved (target unchanged); the new `pageColumnsUpdate?.find(...)` null-safety branch exercised |

### DIFF
--- a/page/index.js
+++ b/page/index.js
@@ -185,37 +185,26 @@ router.route('/:pageId')
               }
             }
           }
-          if (!oldId.startsWith(process.env.RERUMIDPREFIX)) {
+          const columnToUpdate = pageColumnsUpdate?.find(col => col.lines.includes(newId))
+          if (columnToUpdate) {
+            const columnDB = new Column(columnToUpdate.id)
+            const columnData = await columnDB.getColumnData()
             const splitIdsEntry = splitIds.find(pair => Object.keys(pair)[0] === newId)
-            const newColumnRecord = await Column.createNewColumn(pageId, projectId, null, [newId, ...(splitIdsEntry ? splitIdsEntry[newId] : [])])
-            const newColumn = {
-              id: newColumnRecord._id,
-              label: newColumnRecord.label,
-              lines: newColumnRecord.lines
-            }
-            pageColumnsUpdate = pageColumnsUpdate ? [...pageColumnsUpdate, newColumn] : [newColumn]
-          } else {
-            const columnToUpdate = pageColumnsUpdate.find(col => col.lines.includes(newId))
-            if (columnToUpdate) {
-              const columnDB = new Column(columnToUpdate.id)
-              const columnData = await columnDB.getColumnData()
-              const splitIdsEntry = splitIds.find(pair => Object.keys(pair)[0] === newId)
-              if (splitIdsEntry) {
-                const dateIds = splitIdsEntry[newId]
-                columnData.lines.push(...dateIds)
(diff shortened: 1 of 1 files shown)


cubap (approved):
worked as tested

PR #531, opened 2026-06-15
Fix manifest import rejecting valid metadata with non-'none' language keys

### LINKED ISSUE #530: Manifest import rejects valid metadata with non-'none' language keys
## Problem

The manifest import endpoint (`POST /project/import?createFrom=URL`) rejects valid IIIF manifests when metadata values use language tags other than `"none"`.

## Root Cause

In `validatePayload.js`, metadata validation only accepts:
1. Plain string values: `{ value: "John Doe" }`
2. Language maps with `"none"` key: `{ value: { "none": ["John Doe"] } }`

```js
if (typeof metadataItem.value?.none?.[0] !== 'string' && typeof metadataItem.value !== 'string')
  return { isValid: false, errors: 'metadata item value cannot be processed' }
```

Valid IIIF manifests can use **any IETF language tag** (`"en"`, `"fr"`, `"la"`, etc.). A manifest like:
```json
{ "label": { "en": ["Author"] }, "value": { "en": ["John Doe"] } }
```
will be rejected because `value?.none?.[0]` is `undefined` and `value` is an object.

## Impact

Related to [TPEN-interfaces#567](https://github.com/CenterForDigitalHumanities/TPEN-interfaces/issues/567) — users see opaque import errors when importing manifests with language-specific metadata.

## Fix

Relax the validation in `validatePayload.js` to accept any language key in metadata `label` and `value` fields, not just `"none"`. Also ensure `ProjectFactory` properly handles language-mapped metadata values during project creation.

## Files to Check

- `validatePayload.js` — metadata validation logic
- `ProjectFactory.js` — `DBObjectFromManifest()` metadata extraction

cubap:
## Problem

The manifest import endpoint (`POST /project/import?createFrom=URL`) was rejecting valid IIIF manifests when metadata values used language tags other than `"none"`.

## Root Cause

`validatePayload.js` only accepted:

1. Plain string values: `{ value: "John Doe" }`
2. Language maps with `"none"` key: `{ value: { "none": ["John Doe"] } }`

Valid IIIF manifests can use **any IETF language tag** (`"en"`, `"fr"`, `"la"`, etc.).

## Changes

- **`utilities/validatePayload.js`** — Added `validateLanguageMap()` and `validateMetadataField()` helpers to accept any IETF language tag in metadata `label` and `value` fields. Refactored duplicated label/value validation into a single parameterized helper.
- **`test/local/utility.validation.test.js`** — Added 5 test cases covering non-`"none"` language tags, mixed formats, and invalid language maps.

## Testing

All 14 local tests pass, including the 5 new metadata validation tests.

## Related

- Fixes #530
- Related to [TPEN-interfaces#567](https://github.com/CenterForDigitalHumanities/TPEN-interfaces/issues/567)

### DIFF
--- a/test/local/utility.validation.test.js
+++ b/test/local/utility.validation.test.js
@@ -60,4 +60,68 @@ describe('Validation utilities', () => {
     const filteredObject = scrubDefaultRoles({ OWNER: ['*'], CUSTOM_ROLE: ['READ_*_*'] })
     assert.deepEqual(filteredObject, { CUSTOM_ROLE: ['READ_*_*'] }, 'OWNER key must be stripped from a roles map')
   })
+
+  it('accepts metadata with non-"none" language tags', () => {
+    const payload = buildValidProjectPayload()
+    payload.metadata = [
+      { label: { en: ['Author'] }, value: { en: ['John Doe'] } },
+      { label: { fr: ['Auteur'] }, value: { fr: ['Jean Dupont'] } },
+      { label: { la: ['Auctor'] }, value: { la: ['Iohannes'] } }
+    ]
+
+    const result = validateProjectPayload(payload)
+
+    assert.equal(result.isValid, true)
+    assert.equal(result.errors, null)
+  })
+
+  it('accepts metadata with mixed language tags and plain strings', () => {
+    const payload = buildValidProjectPayload()
+    payload.metadata = [
+      { label: 'Source', value: 'Test Source' },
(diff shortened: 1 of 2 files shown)


thehabes (approved):
Made some minor changes during review and testing.  Tested by importing various Manifests with different kinds of valid and invalid `metadata` properties.  I metadata as expected.

